PR #4, opened 2026-04-22
fix: add gracefull shutdown when clicking stop button for proxy server

v4g4b0nd-0x76:
## Fix graceful shutdown for GUI Stop button

**Problem:**
- Clicking Stop in GUI didn't trigger proper shutdown sequence
- Ports remained locked after stopping (address already in use)
- "Shutdown signal received" log never appeared

**Root Cause:**
GUI was immediately aborting the proxy task right after sending shutdown signal, killing it before it could clean up resources.

**Solution:**
- Send shutdown signal first
- Wait up to 2 seconds for graceful shutdown
- Only force-abort on timeout
- Ports now release immediately when stopping

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,7 +97,7 @@ dependencies = [
  "objc2-foundation 0.3.2",
  "parking_lot",
  "percent-encoding",
- "windows-sys 0.52.0",
+ "windows-sys 0.60.2",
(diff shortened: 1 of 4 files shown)


v4g4b0nd-0x76:
sorry if cargo fmt added more changes than required and i know its annoying to review.

PR #6, opened 2026-04-22
chore: Add vertical scrolling to main window for better compatibility…

### LINKED ISSUE #5: Can't Scroll down on the UI version
I'm using the Windows version, and my laptop does not have a tall enough screen to show the whole GUI window. At the same time, there seems to be no scrolling option so almost 1/3 of the bottom of the window is obscured especially after I hit Start.

v4g4b0nd-0x76:
Fixes #5  where GUI content was partially obscured on laptops with short screens.

## Problem

The main window content overflowed on devices with limited vertical space, making some UI elements inaccessible without resizing the window.

## Solution

- Wrapped CentralPanel content in ScrollArea::vertical() with .auto_shrink([false; 2])
- Ensures all UI elements (config form, stats, logs) remain accessible regardless of screen height
- Nested scroll areas (per-site stats, logs) continue to function correctly within the main scrollable area

### DIFF
--- a/src/bin/ui.rs
+++ b/src/bin/ui.rs
@@ -39,7 +39,7 @@ fn main() -> eframe::Result<()> {
     let options = eframe::NativeOptions {
         viewport: egui::ViewportBuilder::default()
             .with_inner_size([WIN_WIDTH, WIN_HEIGHT])
-            .with_min_inner_size([420.0, 540.0])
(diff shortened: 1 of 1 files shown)


therealaleph:
Thanks @v4g4b0nd-0x76 — shipped in v0.7.1 (835a705). Applied the two functional changes (ScrollArea + lower min window height) on top of current main. Co-authored-by trailer is on the commit. 🙏

PR #9, opened 2026-04-22
feat: Add dynamic Google IP discovery with frontend validation

v4g4b0nd-0x76:
## Overview
This PR introduces dynamic IP discovery and validation for the `scan-ips` command, allowing the tool to fetch and test Google IP ranges from the official API instead of relying solely on a static list.

### New Configuration Options
- **`fetch_ips_from_api`** (boolean, default: `true`): Enables dynamic fetching of Google IP ranges from `goog.json` API
- **`max_ips_to_scan`** (usize, default: `100`): Maximum number of candidate IPs to select and test
- **`google_ip_validation`** (boolean, default: `true`): Enables strict frontend validation to ensure IPs serve actual Google services (Search, YouTube, Meet)

### Core Logic Updates
- Resolves IPs from famous Google domains (google.com, youtube.com, etc.)
- Fetches CIDR blocks from Google's public IP ranges API
- Prioritizes IPs from CIDRs containing famous domain IPs
- Randomly selects up to `max_ips_to_scan` candidates (prioritizing famous domain IPs first)
- Validates only the selected candidates instead of all discovered IPs

### Enhanced Validation
- Strict frontend validation checks response headers for:
  - `server: gws`
  - `x-google-*` headers
  - `alt-svc: h3=`
- Ensures discovered IPs are legitimate Google frontend servers, not generic GCP infrastructure

## Important Notes

⚠️ **This is a suggestion and may return wrong results.**

While dynamic IP discovery may find faster IPs than the static fallback list, there is **no guarantee** that discovered IPs will work reliably.

The static fallback list remains available by setting `fetch_ips_from_api: false`.

## Documentation
- Added English and Persian documentation for new configuration options
- Updated README with usage examples and warnings about result reliability

**Note:** This implementation avoids introducing new third-party dependencies to maintain compatibility with the existing release pipeline. The code uses only existing dependencies, which may result in some unconventional patterns.


therealaleph:
Thanks @v4g4b0nd-0x76, merged in v0.8.0. Folded two small fixes into the merge commit: rng was held across an .await (Send bound on the async fn), and /0 CIDR would shift-overflow in the cidr math — both unlikely in practice but defensive. Graceful fallback to the static list works correctly when gstatic fetch fails, verified locally. 🙏

PR #21, opened 2026-04-22
proxy_server: support chunked request bodies

freeinternet865:
Teach the incoming HTTP request parser to handle Transfer-Encoding: chunked instead of only Content-Length-framed bodies.

Also reply with 100 Continue when a client sends Expect: 100-continue before waiting for the request body.

This keeps request framing correct for POST/PUT-style clients and adds focused tests for chunked decoding and 100-continue handling.

-----

Edited and written by GPT.

### DIFF
--- a/src/proxy_server.rs
+++ b/src/proxy_server.rs
@@ -950,7 +950,6 @@ where
         None => return Ok(false),
     };
 
-    // Read body if content-length is set.
     let body = read_body(stream, &leftover, &headers).await?;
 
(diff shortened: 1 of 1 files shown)


therealaleph (approved):
Adds proper chunked request decoding and Expect: 100-continue handling with focused async tests. Also correctly rejects conflicting framing (both chunked and Content-Length) and unsupported transfer encodings. cargo check + tests (51 passing) pass locally.

PR #22, opened 2026-04-22
domain_fronter: drain chunked response trailers

freeinternet865:
Consume the final CRLF and any trailer lines after a zero-length chunk before returning a pooled upstream connection.

Without this, leftover bytes from a chunked response could remain on the socket and corrupt the next response read on the reused connection.

Add a regression test that reads a chunked response with a trailer and then successfully parses a second response from the same stream.

-----

Edited and written by GPT.

### DIFF
--- a/src/domain_fronter.rs
+++ b/src/domain_fronter.rs
@@ -1036,20 +1036,10 @@ where
     let mut out: Vec<u8> = Vec::new();
     let mut tmp = [0u8; 16384];
     loop {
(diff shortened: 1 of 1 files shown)


therealaleph (approved):
Properly drains final CRLF and trailers so a pooled connection is safe to reuse. Clean refactor via read_crlf_line helper, regression test exercises trailer + second response on the same stream. Tests pass locally.

PR #23, opened 2026-04-22
config: reject zero scan batch size

freeinternet865:
Validate scan_batch_size at config load time so scan-ips cannot reach candidate_ips.chunks(0) and panic.

Add a focused regression test for scan_batch_size = 0.

-----

Edited and written by GPT.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -165,6 +165,11 @@ impl Config {
                 ));
             }
         }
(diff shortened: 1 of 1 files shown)


therealaleph (approved):
Small, well-scoped validation with a matching regression test. cargo check + cargo test (lib/bins) pass locally.

PR #24, opened 2026-04-22
domain_fronter: reject invalid relay body base64

freeinternet865:
Treat malformed base64 in Apps Script relay responses as a bad upstream response instead of silently turning it into an empty body.

Add a focused regression test for invalid response body encoding.

-----

Edited and written by GPT.


therealaleph (approved):
Correct fix: malformed base64 should surface as an upstream error instead of becoming a silent empty body. Test added. cargo check + lib/bin tests pass locally.

PR #25, opened 2026-04-22
ui: handle proxy task shutdown more reliably

freeinternet865:
Log server.run() errors in the UI background task instead of silently discarding them.

Also abort the task if graceful shutdown times out, and make sure a finished or stopped proxy task clears the running flag consistently without wiping the last visible stats snapshot.

-----

Edited and written by GPT.


therealaleph (approved):
Surfaces proxy errors that were previously swallowed, forces abort on shutdown timeout, and keeps last_stats visible after stop. cargo check + tests pass locally.

PR #55, opened 2026-04-23
Add optional YouTube relay mode (bypass SNI rewrite)

ghost:
## Summary

  YouTube was going through the same direct Google SNI-rewrite path as the main Google host list. In practice the TLS SNI doesn’t look like a normal YouTube hostname, and some people see weird behavior (restricted
  mode, sign-in quirks, etc.).

  This adds an opt-in: skip_youtube_sni_rewrite. When enabled, the YouTube-related suffixes (youtube.com, youtu.be, ytimg.com, …) are not matched for the SNI-rewrite tunnel and fall back to the Apps Script relay
  instead, so the visible TLS identity matches the site. Tradeoff: likely slower for video, but more predictable for YouTube.

  * **UI** : new checkbox in settings — “Send YouTube through relay (no SNI rewrite)” — with a short tooltip explaining the tradeoff.

  * **Code**: SNI lists split into “core” Google (unchanged) vs YouTube; YouTube is only added to the rewrite path when the new flag is off (default is off, so existing behavior stays the same unless the user turns it on).

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+[target.x86_64-pc-windows-gnu]
+linker = "x86_64-w64-mingw32-gcc"
+
+[target.aarch64-apple-darwin]
+linker = ".cargo/zig-cc-aarch64-apple-darwin.sh"
--- a/.cargo/zig-cc-aarch64-apple-darwin.sh
+++ b/.cargo/zig-cc-aarch64-apple-darwin.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec /opt/zig-linux-x86_64-0.14.0/zig cc -target aarch64-macos "$@"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2186,7 +2186,7 @@ dependencies = [
 
 [[package]]
 name = "mhrv-rs"
-version = "1.1.2"
+version = "1.1.3"
 dependencies = [
  "base64 0.22.1",
  "bytes",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mhrv-rs"
-version = "1.1.3"
+version = "1.1.4"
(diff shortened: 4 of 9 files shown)


Feiabyte:
yes I was gonna make an issue for this. this problem is present in the original version of the script (MasterHttpRelayVPN). if you've managed to fix it and if it gets merged with the project it would be great

therealaleph:
Thanks for the contribution — not merging this one for a couple of reasons:

1. **`.cargo/config.toml` + `zig-cc-aarch64-apple-darwin.sh`**: these are local cross-compile helpers for your machine (`x86_64-w64-mingw32-gcc`, `/opt/zig-linux-x86_64-0.14.0/zig`). Committing them to the repo would break our actual CI cross-compiles, which use a different matrix (dtolnay/rust-toolchain + msys2 for Windows GNU, macos-latest runners for aarch64-apple-darwin). The repo shouldn't carry anyone's per-machine build paths.

2. **Category mix-up**: the PR moves `doubleclick.net`, `googlesyndication.com`, `googleadservices.com`, `google-analytics.com`, `googletagmanager.com`, `googletagservices.com` into `SNI_REWRITE_YOUTUBE_SUFFIXES`. They're ads/analytics, not YouTube — grouping them under a `skip_youtube_sni_rewrite` toggle means anyone who flips the YouTube escape-hatch for video playback reasons also loses direct routing for the ads/analytics domains, which doesn't match the option's description.

3. **Premise**: the TLS SNI is rewritten to the front domain (`www.google.com` by default); the cert served by the GFE covers `*.youtube.com` and the HTTP Host header routes inside. That's how the Python upstream does it too, and we've never seen restricted-mode reports tied to this path. If you have a specific case where YouTube breaks, please open an issue with the logcat/logs — we'll take a look.

Closing — feel free to reopen with the `.cargo/config.toml` bits stripped and the ads/analytics suffixes kept in core.

PR #70, opened 2026-04-23
proxy: fix google_only plain HTTP content length

freeinternet865:
The google_only 502 response for plain HTTP requests advertised Content-Length: 128 even though the body is 120 bytes. Some clients can treat that as a truncated response and wait for bytes that never arrive.

This mismatch was first introduced in https://github.com/therealaleph/MasterHttpRelayVPN-RUST/pull/62.

@dazzling-no-more


freeinternet865:
I pushed the test and then reverted it, so you can evaluate it yourself if needed; however, I believe including it in the final merge would have been unnecessary and disproportionate to the scope of the fix.

PR #120, opened 2026-04-24
Update Code.gs

Mehranzare2020:


therealaleph:
⚠️ **Security: do NOT merge — this PR replaces the placeholder AUTH_KEY in the shared `Code.gs` template with your personal secret.**

The line you changed:
```
-const AUTH_KEY = "CHANGE_ME_TO_A_STRONG_SECRET";
+const AUTH_KEY = "Mehran@#$%1234567bghnncvnncvn";
```

That file ships in every release as a template every user copies into their own Apps Script. The placeholder string is intentional — each user is supposed to change it in **their own local copy** and deploy it to **their own** Apps Script project. Committing your real secret into the shared template leaks your secret to the public (everyone who views this PR can read it, including via git history if merged) AND breaks the template for future users who copy it thinking it's a placeholder.

**Two things to do right now:**

1. **Rotate your AUTH_KEY immediately.** Open your Apps Script at script.google.com, change `AUTH_KEY` in `Code.gs` to a new secret, then Deploy → Manage deployments → Edit → New version → Deploy. Your old key is now publicly readable.
2. Also update `auth_key` in your local `config.json` on whatever device runs mhrv-rs to match the new key.

Closing this PR. No changes needed on our side — the current `CHANGE_ME_TO_A_STRONG_SECRET` placeholder is correct as-is.

---

PR #126, opened 2026-04-24
docs: clarify deployment IDs — same account or multiple accounts

vahidlazio:
## README clarification: deployment IDs

The full tunnel quick-start section previously said "deploy to 3–12 Google accounts", implying you need multiple accounts. In reality you can create multiple deployments on the **same** account (each "New deployment" gets its own ID).

### Changes (English + Farsi)

- Same account deployments: better bandwidth (shared Apps Script instance pool)
- Multiple accounts: more daily quota (each account gets its own 20k calls/day)
- Both approaches work for pipeline depth scaling in full mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ More deployments = more concurrent batches = lower per-session latency. Each bat
 
 ### Quick start
 
-1. Deploy [`CodeFull.gs`](assets/apps_script/CodeFull.gs) to 3–12 Google accounts (same steps as `Code.gs`, but use the full-mode script that forwards to your tunnel-node)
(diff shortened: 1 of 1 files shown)


therealaleph:
Good catch, thanks — the wording *was* ambiguous. Fixed in commit 0bf8ea0 on main (will land in whatever the next tag is, no rush to cut v1.2.14 just for docs).

Rewrote the quick-start step 1 to explicitly say multi-deployment-on-one-account is fine and that multi-account only matters for daily quota. Added three concrete tiers (solo / small group / large group) instead of "3–12 accounts" as if that were the right axis.

One side effect: this also corrects a mistake I made on #61 earlier today — I told @Feiabyte that same-account multi-deployment "doesn't give throughput because per-account concurrency cap", which was wrong in the mhrv-rs use case since we're well below Apps Script's 30-concurrent-executions-per-account ceiling at pipeline_depth ≤ 12. Posting the correction on #61 too.

Closing since this is resolved on main.

---

therealaleph:
Follow-up: you also updated the Persian FAQ entry, which I'd initially missed. Mirrored that into main in ab1d612 with credit to you in the commit body. Both languages now say the same thing.

Thanks for the cleanup — the "3–12 Google accounts" wording was straight-up misleading, and now it's explicit that same-account deployments are the default path.

---

PR #117, opened 2026-04-24
Feature/youtube sni relay

ghost:
https://github.com/therealaleph/MasterHttpRelayVPN-RUST/pull/55
https://github.com/therealaleph/MasterHttpRelayVPN-RUST/pull/115

PR #131, opened 2026-04-24
feat: remove pipeline depth cap — scales with deployment count

vahidlazio:
## Remove artificial pipeline depth cap

Pipeline depth was hard-capped at 12 regardless of how many deployments the user configured. Users with 20+ deployments across multiple accounts were wasting pipeline capacity.

### Change

```rust
// Before
const MAX_PIPELINE_DEPTH: usize = 12;
pipeline_depth = num_scripts.clamp(2, 12);

// After
pipeline_depth = num_scripts.max(2);  // no upper cap
```

The connection pool (POOL_MAX=80) is the natural ceiling — nobody's deploying 80 scripts.

### Updated docs (EN + FA)
- Removed "clamped to 2..12" references
- Added row for 20 deployments in the performance table
- tunnel-node README updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ Full tunnel mode (`"mode": "full"`) routes **all** traffic end-to-end through Ap
 Each Apps Script batch request takes ~2 seconds round-trip. In full mode, `mhrv-rs` runs a **pipelined batch multiplexer** that fires multiple batch requests concurrently without waiting for the previous one to return. The number of in-flight batches (the *pipeline depth*) scales directly with the number of deployment IDs you configure:
 
 ```
-pipeline_depth = number_of_script_ids  (clamped to 2..12)
+pipeline_depth = number_of_script_ids  (minimum 2)
 ```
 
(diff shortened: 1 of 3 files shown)


therealaleph:
Reviewed locally — clean change, behaviorally sound.

**Verified:**
- Merged against current main (post-#117 + v1.2.14), only README.md needed auto-merge, no conflicts
- `cargo test --lib` = 82/82 passes
- Pool ceiling math checks out: `POOL_MAX=80` is the natural upper bound, and nobody's running 80+ deployments in practice
- Docs updated symmetrically — EN + FA both reflect the new "minimum 2, no upper cap" wording, and the table extension to a 20-deployment row is a useful addition for users running multi-account setups
- `tunnel-node/README.md` updated too — nice catch

The cap was always a leftover guess rather than a measured limit, so removing it for users who consciously configure more deployments is the right move. The Apps Script per-account rate limits remain the real ceiling either way, and that's not something we can or should enforce client-side.

Merging. Thanks!

---

PR #142, opened 2026-04-24
feat: per-deployment concurrency (30 req/account)

vahidlazio:
## Summary
- Replace global pipeline semaphore with per-deployment concurrency pools (30 permits each), matching the Apps Script per-account execution limit
- Each deployment ID gets its own semaphore — with N deployments the system sustains 30×N concurrent batch requests instead of just N
- Update README (EN + FA) with new concurrency model and table

## Changes
- `src/tunnel_client.rs`: per-deployment `HashMap<String, Semaphore>` instead of single global semaphore; `fire_batch` picks a deployment, acquires its slot, then fires via `tunnel_batch_request_to`
- `src/domain_fronter.rs`: expose `next_script_id()` and `script_id_list()`; add `tunnel_batch_request_to()` that targets a specific deployment
- `README.md`: updated pipeline docs for both EN and FA sections

## Test plan
- [x] `cargo build` passes
- [x] Local end-to-end test (`mhrv-rs test`) passes with single deployment
- [x] Android APK build + install + launch on device

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ Firefox keeps its own cert store; the installer also drops the CA into Firefox's
 
 Open the UI and fill in the form:
 
-- **Apps Script ID** — the Deployment ID from Step 1. Add multiple IDs (one per line in the UI, or a JSON array in `config.json`) for higher quota **and** lower latency. In `apps_script` mode, IDs are round-robined. In `full` mode, more IDs directly increase the pipeline depth (see [Full tunnel mode](#full-tunnel-mode) below).
+- **Apps Script ID** — the Deployment ID from Step 1. Add multiple IDs (one per line in the UI, or a JSON array in `config.json`) for higher quota **and** lower latency. In `apps_script` mode, IDs are round-robined. In `full` mode, each deployment gets its own pool of 30 concurrent requests (the Apps Script per-account limit), so more IDs = more total throughput (see [Full tunnel mode](#full-tunnel-mode) below).
 - **Auth key** — the same secret you set in `Code.gs`.
 - **Google IP** — `216.239.38.120` is a solid default. Use the **scan** button to probe for a faster one from your network.
 - **Front domain** — keep `www.google.com`.
@@ -271,21 +271,20 @@ Full tunnel mode (`"mode": "full"`) routes **all** traffic end-to-end through Ap
 
 ### How deployment IDs affect performance
 
-Each Apps Script batch request takes ~2 seconds round-trip. In full mode, `mhrv-rs` runs a **pipelined batch multiplexer** that fires multiple batch requests concurrently without waiting for the previous one to return. The number of in-flight batches (the *pipeline depth*) scales directly with the number of deployment IDs you configure:
+Each Apps Script batch request takes ~2 seconds round-trip. In full mode, `mhrv-rs` runs a **pipelined batch multiplexer** that fires multiple batch requests concurrently without waiting for the previous one to return. Each deployment ID (= one Google account) gets its own concurrency pool of **30 in-flight requests** — matching the Apps Script per-account execution limit.
 
 ```
-pipeline_depth = number_of_script_ids  (minimum 2)
+max_concurrent = 30 × number_of_deployment_ids
 ```
 
-| Deployments | Pipeline depth | Effective batch interval | Notes |
-|-------------|---------------|------------------------|-------|
-| 1 | 2 | ~1.0s | Minimum — still pipelines 2 batches |
-| 3 | 3 | ~0.7s | Good for light browsing |
-| 6 | 6 | ~0.3s | Recommended for daily use |
-| 12 | 12 | ~0.17s | Sweet spot for most users |
-| 20 | 20 | ~0.10s | Multi-account setups |
+| Deployments | Concurrent requests | Notes |
+|-------------|-------------------|-------|
+| 1 | 30 | Single account — plenty for light browsing |
+| 3 | 90 | Good for daily use |
+| 6 | 180 | Recommended for heavy use |
+| 12 | 360 | Multi-account power setup |
 
-More deployments = more concurrent batches = lower per-session latency. Each batch round-robins across your deployment IDs, so the load is spread evenly and you're less likely to hit a single deployment's quota ceiling.
+More deployments = more total concurrency = lower per-session latency. Each batch round-robins across your deployment IDs, so the load is spread evenly and you're less likely to hit a single deployment's quota ceiling.
 
 **Resource guards** keep things safe:
 - **50 ops max** per batch — if more sessions are active, the mux splits into multiple batches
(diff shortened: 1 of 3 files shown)


therealaleph:
Thanks for pushing on throughput — but I need to push back on the core premise here before this can merge. There's a footgun hiding in the "one deployment = one account" assumption.

## The issue

Apps Script's "30 simultaneous executions" limit is **per owner account**, not per deployment. From [Google's Apps Script quotas table](https://developers.google.com/apps-script/guides/services/quotas#current_limitations) under Script runtime:

> Simultaneous executions per user — 30 (free)

The "per user" there means the Google user account that owns the script project. If I'm signed in as `alice@gmail.com` and I create 5 deployments of the same script under that account, all 5 deployments share **one** 30-simultaneous-executions bucket. Creating more deployments of the same script in the same account doesn't buy me more concurrency — it just splits the same bucket.

Your code assumes:
```
total_concurrency = 30 × N_deployments
```

But for the majority setup (one user with multiple deployments under their single Google account), the actual behaviour is:
```
total_concurrency = min(30 × N_deployments, 30)    # = 30, always
```

## What breaks in practice

Current code (post-#131): N deployments = N concurrent batches. For the common N=3 single-account case, that's 3 concurrent — well under Google's 30 cap, no errors.

After this PR: N=3 → up to 90 concurrent potential. Google's cap fires somewhere past request 30 with `Service invoked too many times` / `ScriptError: Too many simultaneous invocations`, and the user starts seeing relay failures where they had none before. We'd regress a real case (single-account user) to chase a speculative case (multi-account power user).

## Also in the diff

**Fallback semaphore** at `src/tunnel_client.rs`:
```rust
let sem = sems
    .get(&script_id)
    .cloned()
    .unwrap_or_else(|| Arc::new(Semaphore::new(CONCURRENCY_PER_DEPLOYMENT)));
```

That `unwrap_or_else` silently creates a fresh 30-permit semaphore for any script_id not in the map. If the map somehow lacks an entry (shouldn't happen, but this is the defensive code path), the semaphore becomes useless — every request gets its own private 30-slot pool with no cross-request back-pressure, which amplifies the bug above. Should be a hard `expect` or reuse a single shared fallback.

**README table**: with the new math, 12 deployments = 360 concurrent. Even if 30×N were correct, 360 req / 2s round-trip = 180 req/s sustained. The 20k/day cap is hit in under 2 minutes at that rate. The 12-row number feels worth re-grounding against the realistic ceiling.

**Merge state**: the PR is `CONFLICTING` — needs rebase onto main (post-v1.2.14 + #117 + #131). Minor, but flagging.

## What would unblock this

One of these reframings, any of them would work:

**Option A — Config-time account hint.** Add `deployments_per_account: usize = 1` in config. Per-deployment pool size becomes `30 / deployments_per_account`. Users who stack multiple deployments in one Google account set this to 2, 3, etc. Default 1 preserves the "each deployment = its own account" assumption but makes it explicit + overridable.

**Option B — Safer default.** Drop `CONCURRENCY_PER_DEPLOYMENT` to something like 8–10. 8 × 3 = 24 fits under 30 for the single-account common case. Power users with multi-account setups get 8 × 12 = 96 which is still a 48× improvement over current 12. Misses the theoretical ceiling, hits safely across setups.

**Option C — Keep per-deployment 30 but add a global cap.** `min(sum_of_per_deployment_slots, 30)` unless user sets `multi_account: true`. Preserves the throughput gain for power users who opt in + protects everyone else.

Happy to merge any of these with a clear comment explaining the assumption. As-is, the math is sound only for the multi-account case and actively harmful for the single-account case, so I can't take it.

Also: the claim that this matches "the Apps Script per-account execution limit" in the code comment and README should be corrected — it's the per-*user*/owner-*account* limit, and multiple deployments under one account share it. That framing matters because it's the premise the whole design rests on.

Thanks again for the work — this is a real optimization for the setup you have in mind, and I want to land it in a shape that doesn't backfire for the single-account majority.

---

therealaleph:
Double-checked the premise against official + community sources before landing on my review. Here's the evidence for the "shared per-account, not per-deployment" claim, verbatim from the sources:

**1. [Google's Apps Script quotas table](https://developers.google.com/apps-script/guides/services/quotas)** — two separate caps exist:

> Simultaneous executions per user: 30 / user
> Simultaneous executions per script: 1,000

The 30 is the one that bites here, and it's **per user account**, not per deployment. The per-script 1,000 would only matter if you had 1,000+ different *client* users invoking one web app — not our case, because `Execute as: Me` resolves every invocation to the owner account regardless of who clicked.

**2. [Google's web-app deployment docs](https://developers.google.com/apps-script/guides/web)** — when `Execute as: Me` is set (which it must be for mhrv-rs's relay to work, since the script uses `UrlFetchApp` on behalf of the owner):

> "If your web app runs as the developer (you), everybody shares your quota, instead of each user using their own quota."

All invocations of any deployment of that script count against the owner's per-user 30-bucket.

**3. [Apps Script community consensus](https://groups.google.com/g/google-apps-script-community/c/w38dGO_VuGg)** on the multi-deployment case:

> "Creating multiple deployments with 'execute as me' won't give you separate quotas — all requests still count against your single user quota regardless of which deployment URL is used."

**So the actual math for the common setups:**

| Setup | Per-account cap | What this PR assumes |
|---|---|---|
| N deployments of **one** script in **one** Google account | 30 total (shared) | 30×N ❌ |
| N different scripts in **one** Google account | 30 total (shared) | 30×N ❌ |
| N deployments across **N different** Google accounts | 30×N (separate buckets) | 30×N ✅ |

Only the third case holds — which is a valid power-user setup, but it's not the modal user. And there's nothing in the current code that distinguishes these setups (deployment IDs are opaque — mhrv-rs doesn't know which deployments share an owner).

Happy to merge this once it's reshaped so the default doesn't regress the single-account majority. The three options from my earlier comment still stand; Option A (explicit `deployments_per_account` config knob, default 1) is probably the cleanest — it encodes the assumption honestly, and the user who runs a 4-account fleet just sets the knob to 1 (the default) while the user who stacks 5 deployments in one account sets it to 5 and gets `30 / 5 = 6` per pool.

---

Feiabyte:
wow this turned out to be quite a mess but it seems that @therealaleph got it figured out this seems pretty logical. so basically if we want more speed we have to have more accounts not just deployments right? and every single account should get 30 executions

therealaleph:
Thanks for the fast turnaround. ✅

**What I checked on the force-push:**
- Rebased cleanly against current main — now `MERGEABLE`
- `cargo build --features ui` clean, `cargo test --lib` = 82/82 passes
- README + FA FAQ now honestly state the core assumption ("use one deployment per Google account — multiple deployments on the same account share the pool and don't add concurrency"). This was the blocker from my review; the code behaviour and the documented behaviour are now aligned.

**Merging.**

Two small things to keep in mind for a follow-up (not blockers):

1. The `sems.get(&script_id).cloned().unwrap_or_else(|| Arc::new(Semaphore::new(...)))` fallback is unreachable in practice (same `Vec<String>` feeds both `script_id_list()` and `next_script_id()`), but an `.expect("semaphore must exist for every configured script_id")` would make the invariant clearer to future readers and turn a silent bug into a loud one if someone ever mutates `script_ids` at runtime.

2. The 12 × 30 = 360 row in the table is mathematically correct under the per-account assumption, but at 360 concurrent × 2s RTT the 20,000/day `UrlFetchApp` cap is hit in roughly 2 minutes. Not wrong, just worth a follow-up that adds a "what the daily quota actually lets you sustain" column so users don't size their setup by the concurrency ceiling when the daily cap is the real constraint.

Neither needs to hold this PR. Thanks again.

---

PR #143, opened 2026-04-24
fix(android): app splitting ONLY mode don't mix allowed/disallowed apps

dazzling-no-more:
## Problem

`VpnService.Builder` treats `addAllowedApplication()` and `addDisallowedApplication()` as mutually exclusive — calling both on one Builder makes `establish()` throw `IllegalArgumentException`. The existing code unconditionally calls `addDisallowedApplication(packageName)` (the mandatory self-exclude) before the split-mode `when` block, then calls `addAllowedApplication()` for each selected package in ONLY mode. Result: `establish()` fails silently, the TUN never comes up, and the app either fails to start or regresses to an unrouted state — which manifests to the user as "ONLY mode doesn't work, everything still tunnels (or nothing does)".

The existing code comment at [MhrvVpnService.kt:176-180](https://github.com/therealaleph/MasterHttpRelayVPN-RUST/blob/main/android/app/src/main/java/com/therealaleph/mhrv/MhrvVpnService.kt#L176-L180) already acknowledges the mutual exclusivity — but the implementation doesn't respect it.

## Fix

Move the mandatory self-exclude inside the `when (cfg.splitMode)` block:

- **ALL** / **EXCEPT**: keep calling `addDisallowedApplication(packageName)` — no conflict with other disallowed entries.
- **ONLY**: skip the explicit self-exclude entirely. Self-exclusion is implicit, since our own package is not in the allow-list. Also skip our own `packageName` inside the allow-list loop as a defensive guard against a user selecting mhrv-rs itself.
- **ONLY with empty list**: fall back to the self-exclude so `establish()` still succeeds instead of creating a zero-app VPN.

### DIFF
--- a/android/app/src/main/java/com/therealaleph/mhrv/MhrvVpnService.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/MhrvVpnService.kt
@@ -160,38 +160,39 @@ class MhrvVpnService : VpnService() {
             .addRoute("0.0.0.0", 0)
             .addDnsServer("1.1.1.1")
             .setBlocking(false)
-        try {
-            builder.addDisallowedApplication(packageName)
-        } catch (e: Throwable) {
-            // Shouldn't happen for our own package, but don't hard-fail.
-            Log.w(TAG, "addDisallowedApplication failed: ${e.message}")
-        }
 
-        // Apply user-chosen app splitting on top of the mandatory
-        // self-exclusion above.
+        // Apply user-chosen app splitting. The VpnService API treats
+        // addAllowedApplication and addDisallowedApplication as mutually
+        // exclusive — calling both on one Builder throws
+        // IllegalArgumentException at establish() time, which is the bug
+        // that manifested as "ONLY mode tunnels everything" (establish()
+        // failed silently and the fallback never routed correctly).
         //
-        //   ALL    — no extra restriction; every other app routes through
-        //            us. Matches pre-splitting behaviour.
-        //   ONLY   — allow-list. addAllowedApplication() for each chosen
-        //            package; anything missing from the list bypasses the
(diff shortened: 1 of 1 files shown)


therealaleph:
Reviewed carefully — this is a real bug and the fix is correct.

**The underlying issue verified:** `VpnService.Builder.addAllowedApplication()` and `addDisallowedApplication()` are [explicitly documented as mutually exclusive](https://developer.android.com/reference/android/net/VpnService.Builder#addAllowedApplication(java.lang.String)) on the same Builder — calling both throws `IllegalArgumentException` at `establish()` time. The existing code unconditionally added `addDisallowedApplication(packageName)` before the split-mode switch, which poisoned ONLY mode regardless of what the user selected. The code comment already admitted this in the body but the implementation didn't enforce it.

**Your fix matches the documented intent:**
- **ALL / EXCEPT**: self-exclude via `addDisallowedApplication` — correct, these modes can stack disallows.
- **ONLY (non-empty list)**: skip the self-exclude entirely; we're not in the allow-list so exclusion is implicit. Defensive `if (pkg == packageName) continue` in the allow-list loop is a nice touch — handles the edge case where a user accidentally selects mhrv-rs itself.
- **ONLY (empty list)**: fall back to self-exclude so `establish()` creates a working (if useless) VPN rather than failing with a zero-app allow-list. Correct — the warning log already tells the user what happened.

**Verified locally:**
- Test-merged against current main (post-v1.2.14 + PR #142) — clean auto-merge.
- `./gradlew :app:compileReleaseKotlin` = BUILD SUCCESSFUL.
- Reviewed the final merged file — logic matches the intent described in the PR body.

Merging. Thanks for catching this — the "ONLY mode silently fails" behaviour would have been painful to diagnose from user reports, and it was baked into the current code since splitting was added.

---

PR #150, opened 2026-04-24
fix(android): keep empty ONLY split lists self-excluded

freeinternet865:
When ONLY mode contains only this app or stale packages, no allowed application is successfully added. Android then treats the VPN as applying to every app, which can route more traffic than requested and can also include our own proxy traffic in the TUN path. Count successful addAllowedApplication calls and fall back to the existing ALL-mode self-exclusion behavior when none are usable.

### DIFF
--- a/android/app/src/main/java/com/therealaleph/mhrv/MhrvVpnService.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/MhrvVpnService.kt
@@ -191,11 +191,21 @@ class MhrvVpnService : VpnService() {
                         builder.addDisallowedApplication(packageName)
                     } catch (_: Throwable) {}
                 } else {
+                    var allowed = 0
                     for (pkg in cfg.splitApps) {
                         if (pkg == packageName) continue  // can't tunnel ourselves
-                        try { builder.addAllowedApplication(pkg) } catch (e: Throwable) {
+                        try {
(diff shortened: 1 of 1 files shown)


therealaleph:
Good catch — this is the complement to #143 I should have anticipated. Verified locally:

- Test-merged against current main (post-v1.3.0), clean auto-merge
- `./gradlew :app:compileReleaseKotlin` = BUILD SUCCESSFUL
- Logic matches the intent: counts successful `addAllowedApplication` calls; if zero (all entries were either our own package or uninstalled stale ones), falls back to the ALL-mode `addDisallowedApplication(packageName)` so `establish()` creates a working VPN rather than Android's "empty allow-list means all apps" default behaviour (which would silently route our own proxy traffic through the TUN and loop).

Scenario this fixes:
1. User previously set up ONLY mode with 2 apps
2. User uninstalls both apps
3. `splitApps` still has those two package names in `config.json`, plus possibly `com.therealaleph.mhrv` if selected
4. On Start: all three `addAllowedApplication` calls throw `PackageManager.NameNotFoundException` (we log and skip, per existing code), leaving zero allows registered
5. Before this fix: TUN establishes but Android applies it to every app → including us → loop
6. After: `allowed == 0` check triggers → falls back to self-exclude → TUN routes every app except ours, matching ALL-mode

Merging. Thanks for the follow-up.

---

PR #151, opened 2026-04-24
fix(relay): cap synthetic range stitching size

freeinternet865:
Range-parallel relay trusted the origin's Content-Range total before planning remaining ranges and preallocating the stitched response buffer. A bad upstream could advertise an absurd total and force unbounded memory/CPU work. Bound synthetic stitching to 64 MiB and fall back to a normal single GET for larger totals, preserving correctness without returning a fake truncated 200 response.

### DIFF
--- a/src/domain_fronter.rs
+++ b/src/domain_fronter.rs
@@ -59,6 +59,11 @@ type PooledStream = TlsStream<TcpStream>;
 const POOL_TTL_SECS: u64 = 45;
 const POOL_MAX: usize = 80;
 const REQUEST_TIMEOUT_SECS: u64 = 25;
+const RANGE_PARALLEL_CHUNK_BYTES: u64 = 256 * 1024;
+// Keep synthetic range stitching bounded. Without this, a buggy or hostile
+// origin can advertise `Content-Range: bytes 0-1/<huge>` and make us build a
+// massive range plan or preallocate an enormous response buffer.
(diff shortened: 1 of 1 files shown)


therealaleph:
Good defensive fix. The bug scenario you're protecting against:

1. `relay_parallel_range` probes the origin with `Range: bytes=0-262143` (256 KiB)
2. Origin returns `206 Partial Content` with `Content-Range: bytes 0-262143/9999999999999`
3. Before this PR: we'd plan ~38 million 256 KiB ranges and `Vec::with_capacity(9.9 TiB as usize)` — which would either OOM the process or loop forever fetching chunks that may never complete
4. After: `checked_stitched_range_capacity(total)` returns `None` for totals > 64 MiB, we fall back to `self.relay(method, url, headers, body)` for a single normal GET

64 MiB ceiling is reasonable — Apps Script's per-fetch 50 MiB limit means any single response is capped there anyway, and range-parallel is most useful for files in the 5–50 MiB range. Files over 64 MiB falling back to single GET preserves correctness (the user gets the first 50 MiB and the Apps Script truncation kicks in as usual) and never returns fake success.

Verified locally:
- Test-merged against current main (post-#150 + v1.3.0)
- `cargo test --lib` = 83 tests pass (was 82; 1 new coverage)

Also nice that you extracted `RANGE_PARALLEL_CHUNK_BYTES` to a module-level const — that `256 * 1024` constant appeared in 3 places before; consolidating to one named const is a small readability win.

Merging.

---

PR #153, opened 2026-04-24
feat(tunnel): save one RTT per new HTTPS flow via connect_data op

dazzling-no-more:
## Summary

- Adds a new `connect_data` tunnel op that bundles the client's first bytes (typically the TLS ClientHello) with the connect request. For every new HTTPS flow through the full-mode tunnel, this collapses what used to be two Apps Script round trips (connect → wait → data) into one — saving ~500–2000 ms of user-visible latency per connection on typical deployments.
- The feature spans three independently-deployed components: the Rust client (`src/tunnel_client.rs`), the tunnel-node service (`tunnel-node/src/main.rs`), and the Apps Script forwarder (`assets/apps_script/CodeFull.gs`). Backward compatibility with older deployments of any one layer is a hard requirement — see "Rollout" below.

## How it works

1. After the local CONNECT/SOCKS5 handshake, the client holds the socket open for up to 50 ms (`CLIENT_FIRST_DATA_WAIT`) waiting for the client's first bytes. For TLS, the ClientHello lands well inside that window.
2. If bytes arrive, they're bundled into a `connect_data` op that opens the upstream TCP connection *and* writes the initial payload in one tunnel-node call — no second `data` round trip needed.
3. If nothing arrives within 50 ms (or the port is a known server-speaks-first protocol: 21/22/25/80/110/143/587), we skip the pre-read and fall through to the existing plain `connect` path.

## Rollout & version-skew safety

Because the three components deploy independently, any mismatched pair would hard-break new HTTPS flows without the compatibility work below:

- **Structured error code.** Any updated tunnel-node or Apps Script returns `{e: "...", code: "UNSUPPORTED_OP"}` when it sees an op it doesn't recognize. The client checks `code` first — future refactors of error text can't silently break detection.
- **Legacy string-match fallback.** For deployments predating this change, the client also matches the two legacy error shapes — tunnel-node's `"unknown op: connect_data"` and Apps Script's `"unknown tunnel op: connect_data"`. Covered by tests for each.
- **One-shot fallback cache.** Once any deployment-layer returns unsupported, `TunnelMux` flips an `AtomicBool` and every subsequent session skips the optimistic path entirely — no per-session repeated detection cost.
- **Fallback replay.** When `connect_data` is rejected, the buffered ClientHello is handed to `tunnel_loop` as `pending_client_data` and replayed on the first iteration. No bytes are lost on the fallback. Covered by an integration test that binds a loopback TCP pair and asserts the replayed bytes reach the mux before any socket read.

## Performance & correctness details

- **No head-of-line blocking in batches.** Both `connect` and `connect_data` ops in a batch are dispatched via `tokio::task::JoinSet` on the tunnel-node, so a 10-second TCP connect cannot stall data/close ops landing in the same batch.
- **No double-copy of buffered bytes.** The client carries the initial ClientHello as `Arc<Vec<u8>>` from pre-read through the mux and the potential fallback, avoiding a second 64 KB allocation per session.
- **Session cleanup.** Every exit path from `tunnel_connection` (happy path, write error, bad-base64) explicitly sends `MuxMsg::Close` before returning. Bad-base64 in a response no longer propagates as an io error — it breaks the loop cleanly so Close still fires.
- **Shared validation.** `validate_host_port` is now used by both `handle_connect` and `handle_connect_data_phase1` on the tunnel-node, eliminating copy-paste.

## Observability

A new metric block on `TunnelMux` tracks the four outcomes of the 50 ms pre-read. Every 100 events, an info-level summary is emitted:

```
connect_data preread: 73 win / 12 loss / 14 skip(port) / 1 skip(unsup), mean win time 2830µs (ceiling 50000µs)
```

`mean win time` against the 50 000 µs ceiling tells you whether `CLIENT_FIRST_DATA_WAIT` can be tightened. High `loss` or `skip(unsup)` ratios are the signals that the optimization isn't earning its latency budget for a given deployment.

## Test plan

- [x] `cargo test --lib` — 90 tests pass (was 82; 8 new)
- [x] `cargo test -p mhrv-tunnel-node` — 6 tests pass (all new)
- [x] Structured-code unsupported detection test
- [x] Legacy tunnel-node string detection test (`"unknown op: connect_data"`)
- [x] Legacy Apps Script string detection test (`"unknown tunnel op: connect_data"`)
- [x] Unrelated-error rejection (no false positives on other error shapes)
- [x] Sticky fallback cache — second `mark_connect_data_unsupported` is a no-op
- [x] Port heuristic: 21/22/25/80/110/143/587 skip, 443/8443/853/993 don't
- [x] Pre-read counters track each outcome and accumulate win-time correctly
- [x] `tunnel_loop` replays `pending_client_data` as the first `Data` msg, before any socket read — the critical fallback invariant
- [x] Tunnel-node `handle_connect_data_phase1` opens a session, writes the initial bytes, and returns the session `Arc` for drain
- [x] Tunnel-node `handle_connect_data_single` bundles connect + first bytes end-to-end (verified against an echo server)
- [x] Tunnel-node rejects missing host / bad base64 without leaking a half-open TCP connection (reader task aborted on every error arm)
- [ ] Manual: run client against the deployed tunnel-node and watch for `connect_data preread: ... win / ... loss` summary — verify win ratio on a representative browsing workload
- [ ] Manual: point a client built from this branch at a *pre-change* tunnel-node deployment, confirm `skip(unsup)` counter climbs and HTTPS still works via the fallback path

### DIFF
--- a/assets/apps_script/CodeFull.gs
+++ b/assets/apps_script/CodeFull.gs
@@ -56,6 +56,12 @@ function _doTunnel(req) {
       payload.host = req.h;
       payload.port = req.p;
       break;
+    case "connect_data":
+      payload.op = "connect_data";
+      payload.host = req.h;
+      payload.port = req.p;
+      if (req.d) payload.data = req.d;
+      break;
     case "data":
       payload.op = "data";
       payload.sid = req.sid;
@@ -66,7 +72,10 @@ function _doTunnel(req) {
       payload.sid = req.sid;
       break;
     default:
-      return _json({ e: "unknown tunnel op: " + req.t });
+      // Structured `code` lets the Rust client detect version skew
+      // without substring-matching the error text. Must match
+      // CODE_UNSUPPORTED_OP in tunnel_client.rs and tunnel-node/src/main.rs.
+      return _json({ e: "unknown tunnel op: " + req.t, code: "UNSUPPORTED_OP" });
   }
 
   var resp = UrlFetchApp.fetch(TUNNEL_SERVER_URL + "/tunnel", {
--- a/src/domain_fronter.rs
+++ b/src/domain_fronter.rs
@@ -180,6 +180,11 @@ pub struct TunnelResponse {
     pub eof: Option<bool>,
     #[serde(default)]
     pub e: Option<String>,
+    /// Structured error code from the tunnel-node (e.g. `UNSUPPORTED_OP`).
+    /// `None` for legacy tunnel-nodes; clients should fall back to parsing
+    /// `e` only when this is `None` and compatibility is needed.
+    #[serde(default)]
+    pub code: Option<String>,
 }
 
 /// A single op in a batch tunnel request.
--- a/src/tunnel_client.rs
+++ b/src/tunnel_client.rs
@@ -6,12 +6,13 @@
 //! 30 in-flight requests — matching the per-account Apps Script limit.
 
 use std::collections::HashMap;
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use base64::engine::general_purpose::STANDARD as B64;
 use base64::Engine;
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::io::{AsyncReadExt, AsyncWrite, AsyncWriteExt};
 use tokio::net::TcpStream;
 use tokio::sync::{mpsc, oneshot, Semaphore};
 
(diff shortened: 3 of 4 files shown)


therealaleph:
Substantive review — this is genuinely good work. Merging.

**Reviewed carefully:**

- Test-merged against current main (post-#150 + #151 + v1.3.0) — clean auto-merge
- `cargo test --lib` = **90 tests pass** (was 82 after #151; 8 new for connect_data as stated in your PR body)
- `cd tunnel-node && cargo test` = **6 tests pass** (all new)
- Spot-checked the critical paths:
  - ✅ Structured error code `{e:"...", code:"UNSUPPORTED_OP"}` detection path — future-proof against error-text refactors
  - ✅ Legacy string fallback for both tunnel-node (`"unknown op: connect_data"`) and Apps Script (`"unknown tunnel op: connect_data"`) — covers pre-change deployments
  - ✅ `AtomicBool` sticky fallback cache in `TunnelMux` — no per-session re-detection cost after one "unsupported" response
  - ✅ `pending_client_data` replay in `tunnel_loop` — the bytes-lost-on-fallback invariant is the one that would bite users silently in a version-skew scenario, and it's explicitly tested
  - ✅ Port heuristic (21/22/25/80/110/143/587 skip) — all server-speaks-first protocols, correct list
  - ✅ `JoinSet` per-batch dispatch on the tunnel-node — prevents head-of-line blocking on a slow CONNECT

**What I'm not testing but trusting:**

- Manual items #1 and #2 in your test plan (run client against deployed tunnel-node + version-skew with pre-change deployment) require a running tunnel-node deployment, which I don't have in this session. Your unit tests cover the state transitions that would matter here — mismatch detection, fallback replay, sticky cache — so the automated coverage is the responsible proxy for the manual runs.
- The 50 ms `CLIENT_FIRST_DATA_WAIT` timeout is a fine starting default. The `connect_data preread: X win / Y loss / Z skip(port) / W skip(unsup), mean win time Nµs` summary you added will tell us empirically whether to tighten it.

**Rollout**: this lands on main right now. Users running clients from v1.3.0 (or any previous version) against deployments that include your updated `CodeFull.gs` + tunnel-node get the ~500-2000 ms latency win automatically on every new HTTPS flow. Users running this client against a pre-change deployment hit the fallback path once per session (via the sticky `AtomicBool`), then behave exactly like v1.3.0.

Nice shape on the observability too — `mean win time Nµs (ceiling 50000µs)` is the right framing for "should we tighten the timeout?" rather than just showing the count.

Merging. Thanks for the depth.

---

PR #183, opened 2026-04-25
Feature: udp support

dazzling-no-more:
## Summary

End-to-end UDP support for full-tunnel mode. SOCKS5 client `UDP ASSOCIATE` →
tunnel-mux `udp_open` / `udp_data` ops → tunnel-node UDP sessions → real UDP
to the upstream. QUIC/HTTP3, DNS, and STUN now traverse the tunnel instead
of falling back to TCP or leaking outside it.

Apps Script proxies the new ops opaquely through the existing batch
endpoint — `assets/apps_script/CodeFull.gs` only gets a doc-comment update,
no script-logic change required.

## Architecture

**Proxy (`src/proxy_server.rs`)**

- One per-target `udp_session_task` per active destination, owning the
  tunnel-node sid and serializing both directions onto a single
  `mux.udp_data` call at a time.
- Bounded `mpsc` uplink channel (`UDP_UPLINK_QUEUE = 64`) — `try_send` on
  overflow keeps UDP semantics (drop newest, never block the relay loop).
- Adaptive empty-poll backoff (500 ms → 30 s) so an idle destination
  doesn't burn batch slots.
- Per-`UDP ASSOCIATE` cap (`MAX_UDP_SESSIONS_PER_ASSOCIATE = 256`) with
  FIFO eviction; dispatches tracked in a `JoinSet` so teardown can
  `abort_all()` mid-`mux.udp_data` instead of waiting up to 5 s.
- Source-IP pinned to the control TCP peer; port locked to the **first
  valid (parsed) datagram** — a malformed datagram from the right IP no
  longer DoSes the legitimate flow.
- Datagram size guard (`MAX_UDP_PAYLOAD_BYTES = 9 KB`) — anything larger
  is dropped with a debug log to bound Apps Script quota burn.
- Strict UTF-8 for SOCKS5 ATYP=DOMAIN; non-UTF-8 hostnames are rejected
  at parse rather than forwarded as `U+FFFD` into DNS.

**Tunnel-node (`tunnel-node/src/main.rs`)**

- `UdpSessionInner` with bounded `VecDeque` queue (256 deep, drop-oldest
  on overflow with a `queue_drops` counter).
- `udp_reader_task` surfaces upstream socket errors via an `eof:
  AtomicBool` — propagates as `eof: true` on the next batch drain so
  the proxy task exits instead of polling a zombie session for 120 s.
- `last_active` refreshed only on real activity (uplink send or upstream
  recv); empty long-poll batches do **not** refresh it, so the idle
  reaper can actually fire.
- `handle_batch` Phase 2 drains TCP and UDP independently, both
  event-driven via `wait_for_any_drainable` / `wait_for_any_udp_drainable`
  with the **same** `ACTIVE_DRAIN_DEADLINE` / `STRAGGLER_SETTLE` /
  `LONGPOLL_DEADLINE` knobs introduced in v1.5.0.

**Wire protocol additions (`TunnelResponse.pkts`)**

`Option<Vec<String>>` — base64-encoded UDP datagrams, alongside the
existing `d` field for TCP. `#[serde(default)]` keeps the change
forwards-compatible with older clients/servers.

## Test plan
New tunnel-node tests:

- [x] `udp_open_writes_initial_datagram_and_buffers_reply`
- [x] `udp_queue_overflow_drops_oldest_and_counts`
- [x] `udp_drain_surfaces_upstream_eof`
- [x] `udp_data_for_missing_session_returns_eof`
- [x] `tcp_drain_runs_when_batch_also_contains_udp` (regression for
      mixed TCP/UDP batches)

New proxy parser tests:

- [x] `socks5_udp_ipv4_round_trips`, `socks5_udp_ipv6_round_trips`,
      `socks5_udp_domain_packet_round_trips`
- [x] `socks5_udp_rejects_truncated_inputs`
- [x] `socks5_udp_rejects_fragmented_packets`
- [x] `socks5_udp_rejects_non_utf8_domain`

Manual / staging verification recommended for:

- [ ] Android tun2proxy → SOCKS5 UDP ASSOCIATE → real UDP egress.
      Confirm DNS over UDP, QUIC handshake, and STUN candidate
      gathering all complete via `tunnel-node`.
- [ ] `tunnel-node` UDP session lifecycle: open → datagrams flow →
      `last_active` only bumps on real traffic → 120 s idle reap fires.
- [ ] Teardown: SOCKS5 control disconnect aborts in-flight
      `mux.udp_data` calls; tunnel-node logs `udp session ... closed by
      client` for each.

## Coverage gap

The proxy-side UDP relay (`udp_session_task`, the dispatch loop, and
`UdpRelayState`) is exercised by integration testing in deployment
rather than `cargo test` — direct unit testing would require mocking
`TunnelMux`. Tunnel-node side has unit-test coverage for both the
happy path and the error/eviction/eof paths.

## Compatibility

All changes are additive:

- New `TunnelResponse.pkts` field is `Option<…>` with `#[serde(default)]`.
- New batch ops (`udp_open`, `udp_data`) — older tunnel-nodes return
  `UNSUPPORTED_OP`; clients of older tunnel-nodes are unaffected.
- `TunnelMux::udp_open` / `udp_data` / `close_session` are new public
  methods; nothing existing changed signature.


therealaleph:
Merging — substantial work, well-shaped, additive on the wire. Verified locally:

- Test-merged against current main (post-v1.5.0 + history rewrite) — clean auto-merge, no conflicts.
- `cargo test --lib` = **98 tests pass** (was 92 + 6 new SOCKS5 UDP parser tests, all the ones in your test plan).
- `cargo test -p mhrv-tunnel-node` = **22 tests pass** (was 17 + 5 new UDP session lifecycle tests, including the mixed TCP/UDP batch regression).

**What I checked:**

- Wire-protocol additivity confirmed — `TunnelResponse.pkts: Option<Vec<String>>` with `#[serde(default)]` means a v1.5.x client hitting a v1.6.x tunnel-node sees `pkts=None` (deserialize default) and ignores the field; a v1.6.x client hitting a v1.5.x tunnel-node gets `UNSUPPORTED_OP` on `udp_open` and the existing fallback path handles it. Apps Script middleman is opaque to ops — the doc-comment-only update on `CodeFull.gs` is correct.
- Source-IP pinning + first-valid-datagram port lock: a malformed datagram from the legitimate IP no longer locks the wrong port. The parser tests cover this.
- The 9 KB datagram cap is sensible — QUIC's typical max payload is ~1452 bytes (MTU-1500 minus IP+UDP+QUIC header); DNS truncates well below; STUN messages are tiny. 9 KB gives generous headroom for IPv6 path MTU edge cases without blowing Apps Script quota on rogue traffic.
- `MAX_UDP_SESSIONS_PER_ASSOCIATE = 256` with FIFO eviction is the right shape — most apps churn through a handful of UDP destinations max; 256 is comfortable headroom for browsers spraying QUIC handshakes across CDN endpoints.
- `last_active` only on real activity is the critical bit — without it, idle reaper never fires and zombie sessions accumulate. Test `udp_drain_surfaces_upstream_eof` covers the inverse case (real eof → cleanup), and your `last_active` discipline is documented in the inner type.

**Acknowledged manual-test items:**

The three manual items at the bottom (Android tun2proxy → SOCKS5 UDP, tunnel-node session lifecycle, teardown abort_all) are exactly the operational surface your unit tests don't reach. I'll exercise them after release on the mhrv_test emulator and report back.

Releasing as **v1.6.0** (feature bump). Thanks for the depth and the honest "Coverage gap" note about `udp_session_task` being integration-test territory — that's a fair characterization and the unit tests cover everything that doesn't need a live TunnelMux.

---

PR #187, opened 2026-04-25
fix(android): tighten VPN session lifecycle reliability

dazzling-no-more:
### 1. Connect button could fire while previous Stop was still tearing down
The Compose Connect/Disconnect button was gated on a fixed 2s `transitionCooldown` timer. Worst-case teardown is around 9s (`Tun2proxy.stop()` + 4s tun2proxy join + 5s `rt.shutdown_timeout`), so users could tap Connect after the cooldown expired but before the previous Stop had released the listener socket. `Native.startProxy` then failed with "Address already in use" and the UI looked stuck.

The button now suspends on `VpnState.isRunning` until the service actually flips the flag, with a 12s `withTimeoutOrNull` backstop in case the service fails before flipping (e.g., `establish()` returned null). Side benefit: same gate also debounces the rapid-tap EGL renderer crash the old timer was guarding against.

### 2. `Tun2proxy.stop()` had no timeout
The teardown thread trusted the JNI call to return quickly. Wrapped it in a side thread with a 2s `join()`. If the native call hangs, we log and proceed — the bounded tun2proxy join + bounded `rt.shutdown_timeout` below it still release the listener port.

### 3. fd leak window between `parcelFd.detachFd()` and `Thread.start()`
Once `detachFd()` returned, ownership transferred to `Tun2proxy.run`'s `closeFdOnDrop`. If `Thread.start()` threw (OOM under memory pressure), nobody owned the fd and it leaked for the life of the process. Now wrapped in try/catch; on failure we adopt the int fd into a fresh `ParcelFileDescriptor` purely so we can `close()` it, then the half-started service is torn down cleanly (`stopProxy`, `stopForeground`, `stopSelf`).

### 4. Misleading teardown doc-comment
The doc-block claimed step 2 ("close the TUN fd") forces tun2proxy's `read()` to return EBADF. That call has been a no-op since `detachFd` was added — the real exit mechanism is the cooperative `Tun2proxy.stop()` plus the bounded runtime shutdown. Rewrote the doc-block to call out the bound on each step, and corrected the inline step-2 comment so future debuggers don't chase a phantom safety net.

### 5. Recursive crash trap in MhrvApp's uncaught handler
The default-uncaught-exception handler called `Log.e` unguarded. If logging itself threw (logd dead, OOM in message format), we'd recurse into our own handler with a half-handled original exception. Wrapped the `Log.e` in try/catch so the previous handler still fires with the real failure.

## Files touched
- `android/app/src/main/java/com/therealaleph/mhrv/MhrvApp.kt` — guard `Log.e` in uncaught handler
- `android/app/src/main/java/com/therealaleph/mhrv/MhrvVpnService.kt` — bounded `Tun2proxy.stop()`, fd-leak guard, rewritten teardown doc-block
- `android/app/src/main/java/com/therealaleph/mhrv/ui/HomeScreen.kt` — `awaitingRunning` state-driven button gate

### DIFF
--- a/android/app/src/main/java/com/therealaleph/mhrv/MhrvApp.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/MhrvApp.kt
@@ -42,11 +42,18 @@ class MhrvApp : Application() {
         )
         val previous = Thread.getDefaultUncaughtExceptionHandler()
         Thread.setDefaultUncaughtExceptionHandler { thread, throwable ->
-            Log.e(
-                CRASH_TAG,
-                "uncaught on thread=${thread.name} (id=${thread.id}): ${throwable.message}",
-                throwable,
-            )
+            // Log.e itself can throw on extreme conditions (logd dead,
+            // OOM allocating the formatted message). If we let that
+            // bubble up, we'd recurse into our own handler with a
+            // half-handled original exception; swallow it so the
+            // previous handler still fires with the real failure.
+            try {
+                Log.e(
+                    CRASH_TAG,
+                    "uncaught on thread=${thread.name} (id=${thread.id}): ${throwable.message}",
+                    throwable,
+                )
+            } catch (_: Throwable) { }
             // Let the default handler still terminate the process and
             // show the system "app closed" dialog — we just wanted to
             // get a log line out the door first.
--- a/android/app/src/main/java/com/therealaleph/mhrv/MhrvVpnService.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/MhrvVpnService.kt
@@ -242,13 +242,14 @@ class MhrvVpnService : VpnService() {
         tun = parcelFd
 
         // 3) Start tun2proxy on a worker thread. It blocks until stop() or
-        //    shutdown. We detach the fd so ownership transfers cleanly; the
-        //    ParcelFileDescriptor (`tun`) still holds a reference, so closing
-        //    it at teardown reliably tears down the TUN even if tun2proxy
-        //    doesn't cleanly exit.
+        //    shutdown. We detach the fd so ownership transfers cleanly to
+        //    tun2proxy (closeFdOnDrop = true closes it on return from run()).
+        //    The ParcelFileDescriptor (`tun`) we keep is post-detach — its
+        //    own close() is a no-op for the underlying fd, so the worker is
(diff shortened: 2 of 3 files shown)


therealaleph:
Reviewed locally — five clean reliability fixes, low blast radius (Android-only), good rationale on each. Merging.

**Verified:**
- Test-merged against current main (post-v1.6.0 + UDP support) — clean auto-merge.
- `./gradlew :app:compileReleaseKotlin` = BUILD SUCCESSFUL.

**Each fix makes sense:**

1. **Connect button vs. teardown race** — the 2s `transitionCooldown` was a guess; switching to a state-driven `VpnState.isRunning` gate with a 12s `withTimeoutOrNull` backstop is correct because the actual teardown ceiling (Tun2proxy + tun2proxy join + rt.shutdown_timeout = ~9s worst-case) is what we need to wait on. The same gate also gives us the EGL crash debounce as a side effect.

2. **`Tun2proxy.stop()` unbounded** — wrapping in a 2s `join()` is the right shape; the native call is supposed to return fast but should never be allowed to hang the teardown thread. If it does hang, the bounded tun2proxy thread join + bounded `rt.shutdown_timeout` below still release the listener port. We accept the fd-leak as the cost of proceeding.

3. **fd leak between `detachFd()` and `Thread.start()`** — this is a real correctness bug, not just defensive cleanup. Once `detachFd` returns, the ParcelFileDescriptor no longer owns the fd; if `Thread.start` throws (OOM under memory pressure is plausible on cheap Android devices), nobody closes it. Adopting the raw int back into a fresh `ParcelFileDescriptor` for the express purpose of closing it is a tidy idiom — Android FileDescriptor doesn't expose a portable close otherwise.

4. **Misleading doc comment** — fair callout. The "step 2: close TUN fd to force EBADF on read" note has been factually wrong since `detachFd` landed; nobody noticed because the cooperative `Tun2proxy.stop()` has been doing the work. Worth fixing the comment so future debuggers don't waste time on the phantom path.

5. **Recursive crash trap in uncaught handler** — small but real. If `Log.e` itself throws (logd dead, OOM during string format), the default handler never sees the original exception and we lose the actual crash signal. Guarding it with try/catch and falling through to the previous handler is the correct shape.

Releasing as **v1.6.1** (patch — all fixes, no new behavior).

---

PR #245, opened 2026-04-26
fix: add twitter.com

Parsa307:
Hi there, I added twitter.com there is a reason if you use a extension that has the option to redirect to twitter.com like Control Panel for Twitter it will not do the shortening.
Screenshot_20260426_105846_Fennec


Parsa307:
@therealaleph, LGTM.

PR #266, opened 2026-04-26
feat(android): config import/export — clipboard, QR, deep link, share

yyoyoian-pixel:
## Summary

One-tap config sharing between devices. Export your config as a QR code or compressed link, import by pasting, scanning, or tapping a deep link.

## Features

**Import:**
- **Clipboard paste** — banner appears at the top when `mhrv://...` or raw JSON is detected in clipboard. One tap to import, clipboard cleared after.
- **QR scanner** — opens camera in portrait, scans and imports.
- **Deep link** — tapping `mhrv://...` in any app/chat auto-opens mhrv-rs and imports.

**Export:**
- **Unified dialog** — QR code + compressed hash + copy button in one view.
- **Share sheet** — sends QR image + text hash together via Android share (Telegram, WhatsApp, email, etc.).
- **Compact encoding** — only non-default fields, DEFLATE compressed before base64. Typical config is ~200 chars vs ~800 raw.

**Security:**
- Warning: "This includes your auth_key. Only share with people you trust."

## What's shared vs not

| Included | Not included (device-specific) |
|----------|-------------------------------|
| mode, script_ids, auth_key | listen_host, listen_port, socks5_port |
| google_ip (if changed), front_domain (if changed) | connectionMode (VPN/proxy) |
| sni_hosts, verify_ssl, log_level, parallel_relay | splitMode, splitApps |
| upstream_socks5, passthrough_hosts | uiLang |

## Changes

| File | What |
|------|------|
| `ui/ConfigSharing.kt` (new) | Import/export composables, QR generation, scanner launcher |
| `ConfigStore.kt` | `encode()`, `decode()`, `looksLikeConfig()`, `loadFromJson()` (deduplicated from `load()`) |
| `MainActivity.kt` | Deep link intent handling (`mhrv://` scheme) |
| `AndroidManifest.xml` | Deep link intent filter, FileProvider for QR sharing, ZXing portrait lock |
| `build.gradle.kts` | ZXing core + zxing-android-embedded dependencies |
| `HomeScreen.kt` | `ConfigSharingBar` wired in above Mode dropdown |
| `strings.xml` | 14 new string resources |
| `xml/file_paths.xml` (new) | FileProvider paths for cache dir |

## Test plan
- [x] Export → copy → paste on another device → import works
- [x] Export → share via Telegram → recipient pastes → import works
- [x] QR code scans successfully from another phone's screen
- [x] Deep link `mhrv://...` opens app and imports
- [x] Clipboard banner disappears after import (clipboard cleared)
- [x] Raw JSON in clipboard also detected and importable
- [x] QR scanner opens in portrait

### DIFF
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -136,6 +136,10 @@ dependencies {
     implementation("androidx.compose.material3:material3")
     implementation("androidx.compose.material:material-icons-extended")
 
+    // QR code generation + scanning (self-contained, no ML Kit needed).
+    implementation("com.google.zxing:core:3.5.3")
+    implementation("com.journeyapps:zxing-android-embedded:4.3.0")
+
     debugImplementation("androidx.compose.ui:ui-tooling")
     debugImplementation("androidx.compose.ui:ui-test-manifest")
 }
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -59,7 +59,7 @@
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
-                <data android:scheme="mhrv" />
+                <data android:scheme="mhrv-rs" />
             </intent-filter>
         </activity>
 
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -53,7 +53,32 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <!-- Deep link: tapping mhrv://... in any app opens MainActivity
+                 and auto-imports the encoded config. -->
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="mhrv" />
+            </intent-filter>
         </activity>
+
+        <!-- FileProvider for sharing QR code images via the share sheet. -->
+        <provider
+            android:name="androidx.core.content.FileProvider"
+            android:authorities="${applicationId}.fileprovider"
+            android:exported="false"
+            android:grantUriPermissions="true">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/file_paths" />
+        </provider>
+
+        <!-- Force ZXing scanner to portrait (matches app orientation). -->
+        <activity
+            android:name="com.journeyapps.barcodescanner.CaptureActivity"
+            android:screenOrientation="portrait"
+            tools:replace="android:screenOrientation" />
 
         <!--
           VpnService: Android captures all traffic at the IP layer and feeds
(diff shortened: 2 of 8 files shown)


therealaleph (changes requested):
Hi @yyoyoian-pixel — feature is well-built (clipboard banner + QR + DEFLATE compression + careful field-by-field include/exclude list is exactly right). One blocker on security grounds before this can merge:

**🚨 Blocker — `handleDeepLink` auto-imports without confirmation. The QR scanner path is safe; the deep link path is the inconsistency:**

`MainActivity.kt:`
```kotlin
private fun handleDeepLink(intent: Intent?) {
    val data = intent?.data ?: return
    if (data.scheme != "mhrv") return
    val encoded = data.toString()
    val cfg = ConfigStore.decode(encoded) ?: return
    ConfigStore.save(this, cfg)                    // ← overwrites existing config
    android.widget.Toast.makeText(this, "Config imported", ...).show()
}
```

vs. `ConfigSharing.kt` scanner path (correct):
```kotlin
val scanLauncher = rememberLauncherForActivityResult(ScanContract()) { result ->
    val decoded = ConfigStore.decode(scanned)
    if (decoded != null) {
        pendingImport = decoded
        showImportConfirm = true   // ← user has to confirm
    }
}
```

**Why this matters for our threat model:**

This project's user base is exactly the population an attacker would want to MITM — Iranian users who can't use a regular VPN, who often discover tools via Telegram/forum links, who tap things without scrutiny. The attack flow:

1. Attacker crafts `mhrv://<base64-of-attacker-config>` — their own `script_ids` + `auth_key`
2. Posts the link in a Telegram channel claiming "free fast deployment"
3. User taps → mhrv-rs opens → config silently overwrites → traffic now flows through attacker's Apps Script
4. Attacker has full visibility into URLs visited and can MITM HTTPS (mhrv-rs's MITM CA is already trusted by the device, and the CA private key is in the attacker-controlled deployment indirectly via the relay path — wait, actually the CA key stays local; but the attacker doesn't need to break TLS, they just see all the destination URLs and can return arbitrary responses)

The Toast only appears AFTER the config has already been overwritten. By the time the user reads it, they're already routing traffic through the attacker.

**The fix is small** — make `handleDeepLink` use the same `showImportConfirm` modal the scanner path does:

```kotlin
private fun handleDeepLink(intent: Intent?) {
    val data = intent?.data ?: return
    if (data.scheme != "mhrv") return
    val encoded = data.toString()
    val cfg = ConfigStore.decode(encoded) ?: return
    // Don't auto-save. Stash for the UI to confirm.
    pendingDeepLinkImport = cfg
}
```

And then in `AppRoot()` / `HomeScreen`, surface a `pendingDeepLinkImport`-driven dialog using the same `showImportConfirm` UX — show the new `script_ids[0..3]`, the `mode`, and the `tunnel_node_url` (if Full mode), with "Import (replaces current config)" and "Cancel" buttons. The dialog should explicitly note that imported deployment IDs and auth_key will route the user's traffic.

**The clipboard banner path is fine** because the user has to tap "Import" — that's already explicit consent.

**Other points (non-blocking, suggestions):**

- The export warning is good ("This includes your auth_key. Only share with people you trust.") — same warning text on the import side would be appropriate, with the inverse phrasing: "Importing routes your traffic through the deployment IDs in this config. Only import from trusted sources."
- Consider adding a fingerprint display on the import-confirm dialog: hash of `(script_ids[0], auth_key[..8])` shown as 4-byte hex, so users can verify with the sender out-of-band that they got the right config (not a swapped one).
- The `mhrv://` scheme is fairly generic — collision risk with any future app using the same scheme. `mhrv-rs://` would be slightly more specific. Not blocking but worth thinking about.

The clipboard / QR / share / compact-encoding parts are excellent. Once the deep-link auto-import becomes confirm-first, this is mergeable. Happy to re-review on push.

---

yyoyoian-pixel:
@therealaleph fixed, please check if all is fine.

Issue #263, opened 2026-04-26
[Performance/Enhancement] High Latency and Timeouts in "mode: full" via Google App Script (GAS)


w0l4i:
### 🚀 Enhancement Request: Optimizing "mode: full" Performance and Stability in Large-Scale GAS Deployments

I am using `MasterHttpRelayVPN` in `mode: full` with a multi-account architecture to bypass restrictions. My setup is as follows:
- **Architecture:** `Client > 7x Google Apps Script (GAS) Deployments (using 7 distinct Google accounts) > Node Nginx > Tunnel Container`.
- **Concurrency:** Utilizing parallel dispatching with a factor of 2-3 across these 7 deployments.

### 📉 The Problem: Bottleneck in Client Dispatching
Despite distributing the load across 7 independent GAS deployments, I am observing high latency and frequent connection timeouts. The current implementation of the Rust client seems to struggle with managing these concurrent requests efficiently, particularly when the network or Google's endpoints face momentary congestion.

The logs consistently show long RTT spikes leading to `os error 10060`. Instead of failing over to other healthy GAS deployments, the client appears to wait for the stalled request to time out, which creates a "blocking" effect that degrades the entire tunnel's performance.

**Log evidence:**
```text
2026-04-26T15:20:32.688953Z  INFO batch: 1 ops → AKfycbwA, rtt=21.0450168s
2026-04-26T15:20:32.688974Z  WARN batch failed: io: A connection attempt failed ... (os error 10060)
2026-04-26T15:20:34.877856Z  INFO batch: 1 ops → AKfycbwA, rtt=21.0389911s
2026-04-26T15:20:34.877877Z  WARN batch failed: io: A connection attempt failed ... (os error 10060)

💡 Suggested Optimizations
To effectively utilize parallel deployments and resolve these bottlenecks, I suggest the following refactorings:

1. "Fail-Fast" Dispatcher & Timeout Logic
Waiting 21 seconds for a batch to fail is unacceptable for real-time traffic.

Proposal: Implement a configurable request_timeout (e.g., 5-8 seconds). If a request to a specific GAS Deployment ID exceeds this limit, the client should treat it as a failure, mark that deployment as "temporarily degraded," and offload the batch to the next available deployment immediately.

2. Deployment Health-Scoring System
With 7 deployments in use, the client needs better awareness of backend health.

[...]

PR #267, opened 2026-04-26
Add batching configuration options

w0l4i:
Added new configuration o## Overview
This PR addresses the performance bottlenecks identified in issue #263.  As discussed, it moves hardcoded constants related to batching and timeouts into the configuration file.

## Changes
- Added `batch_timeout_ms` to config (default: 10000ms).
- Added `batch_coalesce_window_ms` to config (default: 8ms).
- Added `max_batch_ops` to config (default: 50).ptions for batching with defaults. ## Impact
Users in high-latency environments (like GAS relays) can now tune these parameters to prevent long-tail blocking and improve tunnel responsiveness.

PR #269, opened 2026-04-26
Update tunnel_client.rs

w0l4i:
## Overview
This PR addresses the performance bottlenecks identified in issue #263. 
As discussed, it moves hardcoded constants related to batching and timeouts into the configuration file.

## Changes
- Added `batch_timeout_ms` to config (default: 10000ms).
- Added `batch_coalesce_window_ms` to config (default: 8ms).
- Added `max_batch_ops` to config (default: 50).
- Refactored `tunnel_client.rs` to use these configurable values instead of static constants.

## Impact
Users in high-latency environments (like GAS relays) can now tune these parameters to prevent long-tail blocking and improve tunnel responsiveness.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -78,6 +78,12 @@ pub struct Config {
     pub hosts: HashMap<String, String>,
     #[serde(default)]
     pub enable_batching: bool,
+    #[serde(default = "default_batch_timeout_ms")]
+    pub batch_timeout_ms: u64,
+    #[serde(default = "default_batch_coalesce_window_ms")]
+    pub batch_coalesce_window_ms: u64,
+    #[serde(default = "default_max_batch_ops")]
+    pub max_batch_ops: usize,
     /// Optional upstream SOCKS5 proxy for non-HTTP / raw-TCP traffic
     /// (e.g. `"127.0.0.1:50529"` pointing at a local xray / v2ray instance).
     /// When set, the SOCKS5 listener forwards raw-TCP flows through it
@@ -164,7 +170,9 @@ pub struct Config {
     #[serde(default)]
     pub passthrough_hosts: Vec<String>,
 }
-
+fn default_batch_timeout_ms() -> u64 { 10000 }
+fn default_batch_coalesce_window_ms() -> u64 { 8 }
+fn default_max_batch_ops() -> usize { 50 }
 fn default_fetch_ips_from_api() -> bool { false }
 fn default_max_ips_to_scan() -> usize { 100 }
 fn default_scan_batch_size() -> usize {500}
@@ -219,6 +227,12 @@ impl Config {
                     ));
                 }
             }
+        }
+        if self.batch_timeout_ms == 0 {
+            return Err(ConfigError::Invalid("batch_timeout_ms must be greater than 0".into()));
+        }
+        if self.max_batch_ops == 0 {
+            return Err(ConfigError::Invalid("max_batch_ops must be greater than 0".into()));
         }
         if self.scan_batch_size == 0 {
             return Err(ConfigError::Invalid(
--- a/src/tunnel_client.rs
+++ b/src/tunnel_client.rs
@@ -41,19 +41,19 @@ const MAX_BATCH_OPS: usize = 50;
 /// Timeout for a single batch HTTP round-trip. If the tunnel-node or Apps
 /// Script takes longer than this, the batch fails and sessions get error
 /// replies rather than hanging forever.
-const BATCH_TIMEOUT: Duration = Duration::from_secs(30);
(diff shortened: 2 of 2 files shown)


therealaleph:
Closing this iteration; the intent is correct but the patch doesn't compile yet. When I checked it out and ran \`cargo build --release\`, I got 7 errors:

\`\`\`
error[E0425]: cannot find value \`BATCH_TIMEOUT\` in this scope
   --> src/tunnel_client.rs:610:71
error[E0425]: cannot find value \`config\` in this scope
   --> src/tunnel_client.rs:645:61
error[E0425]: cannot find value \`REPLY_TIMEOUT\` in this scope
   --> src/tunnel_client.rs:926:47
\`\`\`

What's blocking it:

1. **Constants commented out as \`/// const ...\`** — that turns them into doc comments attached to the next item, but the code at lines 610, 645, 926 still references the names. Either delete the lines fully, or keep the consts but make them \`pub(crate)\` defaults that the config field overrides. The doc-comment form is the wrong shape.
2. **\`config\` isn't in scope** in \`mux_loop\` / \`fire_batch\` / \`tunnel_connection\` — those functions don't take a Config parameter today. To wire the new fields you'll need to plumb \`Arc<Config>\` (or just the three timing values) down through the call chain. That's the larger refactor, and the bulk of why I split this into "Pass 1 (config fields)" and "Pass 2 (use them)" originally.
3. **\`config.client_first_data_wait_ms\` is referenced** but never added to \`Config\` — the patch updates the call site but not the struct, so even after fixing (1) and (2), this would still fail.

This is a real architectural change, not a 15-line tweak. **Two paths forward:**

**A — I'll do Pass 1 myself.** Take the field shape you proposed in #267 (which is right), plumb \`Arc<Config>\` down through \`TunnelMux::start\` → \`mux_loop\` → \`fire_batch\`, ship as v1.7.3. You stay credited as the design contributor. This unblocks Pass 2 (the per-deployment health tracking from #263) which has more design surface.

**B — You take another pass.** With the three issues above resolved + a local \`cargo build --release\` verifying it compiles. I'll review the moment it's pushed.

I'd take A — it's a few hours of work and the longer-tail Pass 2 is more interesting use of your time. But B is fine if you'd rather drive both. Let me know which.

Either way, neither #267 nor this PR can land in their current shape. Closing as superseded; re-open with the fix or wait for v1.7.3 with my version.

---

w0l4i:
> Closing this iteration; the intent is correct but the patch doesn't compile yet. When I checked it out and ran `cargo build --release`, I got 7 errors:
> 
> ``` error[E0425]: cannot find value `BATCH_TIMEOUT` in this scope --> src/tunnel_client.rs:610:71 error[E0425]: cannot find value `config` in this scope --> src/tunnel_client.rs:645:61 error[E0425]: cannot find value `REPLY_TIMEOUT` in this scope --> src/tunnel_client.rs:926:47 ```
> 
> What's blocking it:
> 
> 1. **Constants commented out as `/// const ...`** — that turns them into doc comments attached to the next item, but the code at lines 610, 645, 926 still references the names. Either delete the lines fully, or keep the consts but make them `pub(crate)` defaults that the config field overrides. The doc-comment form is the wrong shape.
> 2. **`config` isn't in scope** in `mux_loop` / `fire_batch` / `tunnel_connection` — those functions don't take a Config parameter today. To wire the new fields you'll need to plumb `Arc` (or just the three timing values) down through the call chain. That's the larger refactor, and the bulk of why I split this into "Pass 1 (config fields)" and "Pass 2 (use them)" originally.
> 3. **`config.client_first_data_wait_ms` is referenced** but never added to `Config` — the patch updates the call site but not the struct, so even after fixing (1) and (2), this would still fail.
> 
> This is a real architectural change, not a 15-line tweak. **Two paths forward:**
> 
> **A — I'll do Pass 1 myself.** Take the field shape you proposed in #267 (which is right), plumb `Arc` down through `TunnelMux::start` → `mux_loop` → `fire_batch`, ship as v1.7.3. You stay credited as the design contributor. This unblocks Pass 2 (the per-deployment health tracking from #263) which has more design surface.
> 
> **B — You take another pass.** With the three issues above resolved + a local `cargo build --release` verifying it compiles. I'll review the moment it's pushed.
> 
> I'd take A — it's a few hours of work and the longer-tail Pass 2 is more interesting use of your time. But B is fine if you'd rather drive both. Let me know which.
> 
> Either way, neither #267 nor this PR can land in their current shape. Closing as superseded; re-open with the fix or wait for v1.7.3 with my version.
> 
> [reply via Anthropic Claude | reviewed by @therealaleph]

i will wait for your turn , I'm sorry without good internet i can't run test build and this happened please do according to the plan by yourself #263

Issue #313, opened 2026-04-27
App script Unresponsive ERROR


Surenabedimiran:
Approximately from 12:25 (Tehran local time), fetch requests to Apps Script have become unresponsive. Can you determine whether it's Google blocking the domain-fronting, or Iran’s DPI is somehow reading the HTTP Host? I think Google has blocked the method because bootstrap mode works fine.

therealaleph:
@Surenabedimiran — important signal. Let me try to narrow down whether it's Google-side or Iran-side, and what we can do about it.

**Key observation:** bootstrap (`google_only`) still works = the SNI-rewrite tunnel to `google_ip` with `SNI=www.google.com` is fine. So at the TLS level, traffic to Google edge IPs is unaffected. The break is **specifically on Apps Script execution** — i.e. when Apps Script tries to `UrlFetchApp.fetch(...)` outbound from Google's datacenter.

That narrows it considerably:

**1. Iran DPI reading Host header? → No, not possible.**
After SNI rewrite, the inner HTTP request body (with `Host: script.google.com`) is encrypted end-to-end with the cert Google edge presents (`www.google.com`'s cert). Iran's DPI sees only the encrypted bytes between user ↔ google_ip — it can't read the inner `Host`. If DPI was the issue, `google_only` would also fail.

**2. Google blocking domain-fronting at Apps Script layer? → Plausible, this is the most likely explanation.**
Google's Apps Script gateway can read `Host: script.google.com` (it's plaintext HTTP-after-TLS-termination at Google's edge). They could check whether the originating SNI matches the Host. If SNI was `www.google.com` but Host is `script.google.com`, that's the domain-fronting fingerprint. Google deprecating this server-side has been forecasted for years (Cloudflare/AWS already did, Google was the holdout).

**3. Iran-specific Apps Script throttle? → Less likely but possible.**
Google might be throttling Apps Script requests originating from IP ranges that look like Iran-origin. Bootstrap (google_only) doesn't trip this because it's not running an Apps Script — it's just SNI-rewrite to Google edge for any URL.

**Diagnostic to distinguish:**

Can you ask anyone outside Iran to test the same Apps Script deployment from their region?

```bash
# Outside Iran:
curl -X POST "https://script.google.com/macros/s/YOUR_DEPLOYMENT/exec" \
  -H "Content-Type: application/json" \
  -d '{"k":"YOUR_KEY","u":"https://httpbin.org/get","m":"GET"}' \
[...]

PR #319, opened 2026-04-27
feat(tunnel-client): blacklist deployments after sustained timeouts

dazzling-no-more:
A deployment that returns 3 read-timeouts (or whole-batch timeouts) within any 30-second window gets blacklisted for 120 s. The strike counter resets on any successful batch, so unrelated transient slowness over time can't accumulate. next_script_id() already skips blacklisted deployments, so round-robin will route around the bad one until cooldown expires — at which point the deployment is re-probed naturally and rejoins if it's healthy.

For your earlier scenario (5/8 deployments stale): once any session sends 3 batches to a stale deployment within 30s, it drops out for 2 minutes. New sessions then have a (3 + recovering) / 8 chance of landing on a working deployment — much better than the prior 3/8. If staleness persists, deployments cycle in and out of the cooldown but never serve a critical mass of bursty traffic.

### DIFF
--- a/src/domain_fronter.rs
+++ b/src/domain_fronter.rs
@@ -102,6 +102,13 @@ pub struct DomainFronter {
     inflight: Arc<Mutex<HashMap<String, broadcast::Sender<Vec<u8>>>>>,
     coalesced: AtomicU64,
     blacklist: Arc<std::sync::Mutex<HashMap<String, Instant>>>,
+    /// Per-deployment rolling timeout counter. Maps `script_id` →
+    /// `(window_start, strike_count)`. Reset when the window expires
+    /// or when a batch succeeds. Triggers a short-cooldown blacklist
+    /// at `TIMEOUT_STRIKE_LIMIT`. Distinct from `blacklist` because
+    /// strike state is per-deployment health bookkeeping, not the
+    /// permanent ban list.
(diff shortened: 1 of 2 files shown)


therealaleph (approved):

Approving and merging — exactly the right architectural fit for the #313 Iran outage scenario as well as #305/#307 (sustained timeouts on deployments that look healthy at config-load time but stall mid-batch).

Verified:
- `FronterError::Timeout` exists (src/domain_fronter.rs:53), matches the per-read deadline path
- `next_script_id()` already filters by blacklist (line 366) so round-robin transparently routes around the cooldown'd deployment
- Strike window math is correct — first strike inserts `(now, 1)` via `or_insert((now, 0))` + `entry.1 += 1`; subsequent strikes within the window increment, stale window resets to `(now, 1)`
- `record_batch_success` clears strikes on every successful batch — no cross-incident accumulation
- 120s timeout-cooldown vs 600s quota-cooldown distinction is correct: a stalled deployment is a transient signal, a quota error is sticky for a refill cycle

The two-tier blacklist (permanent ban list vs strike-state bookkeeping) is cleaner than the unified `DeploymentHealth` struct I had in the v1.8.x roadmap. Adopting this design instead.

Merging now. Will cut v1.7.8 once Android emulator smoke-test passes.

Surenabedimiran:
But if the theory is correct, Google still blocks the outbounds right?

PR #334, opened 2026-04-27
Rename README.md to aREADME.md

Mrl423:


therealaleph:
Closing — this rename has no purpose and would break every link to README.md across docs, releases, and external mirrors. If there's a real motivation (e.g. you encountered a sort-order issue), please open an issue describing it so we can solve the actual problem.

PR #337, opened 2026-04-27
Fix Apps Script decoded range probe handling

freeinternet865:
Accept a synthetic first range probe when Content-Range proves the whole entity was returned, even if Apps Script decoded the body and left compressed Content-Range metadata intact. The response is then rewritten to HTTP 200 with Content-Range removed and Content-Length based on the decoded body, avoiding an unnecessary fallback full GET.

Keep strict validation for real client Range requests and later chunks. Also recognize localized Apps Script bandwidth quota errors.

### DIFF
--- a/src/domain_fronter.rs
+++ b/src/domain_fronter.rs
@@ -715,9 +715,9 @@ impl DomainFronter {
     ///      by relay() already (we skip cache for it).
     ///   2. Probe with `Range: bytes=0-<chunk-1>`.
     ///   3. 200 back (origin doesn't support ranges) → return as-is.
-    ///   4. 206 back → parse Content-Range total. If the body fits in
-    ///      the first probe (total <= chunk or body >= total), rewrite
-    ///      the 206 to a 200 so the client — which never asked for a
+    ///   4. 206 back → parse Content-Range total. If Content-Range says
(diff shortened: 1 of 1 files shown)


therealaleph (approved):

Approving and merging — this is the right shape for the bug.

Verified:
- Only touches `src/domain_fronter.rs` (no Cargo.toml/lock changes, no supply-chain pattern)
- The carve-out is properly scoped: `probe_range_covers_complete_entity` only applies to the synthetic first probe (where we control the request and `requested_end`); later chunks still go through the strict `content_range_matches_body` validation
- The added quota-error string matches (German `bandbreitenkontingent`, `datenübertragungsrate`, generic `bandwidth`/`transfer rate`/`limit exceeded`) are sensible — Apps Script's quota errors are localized depending on the account's Google preferred language, and we were silently retrying on responses we should have been blacklisting
- Tests added in `mod tests` covering the new branch

The decoded-body case (Apps Script returns the gzip-decoded entity but echoes the origin's compressed `Content-Range`) was a real source of unnecessary fallback full GETs in the wild — this should reduce wasted Apps Script quota for sites that 206 with gzip.

Nice surgical fix, thanks @freeinternet865.

PR #344, opened 2026-04-27
fix(android): replace broken clipboard auto-detect with Paste button

yyoyoian-pixel:
## Bug

The "Config detected in clipboard" banner was intermittently not showing — sometimes it appeared on first app open but never on resume from another app. This made config import via clipboard unreliable.

**Root cause**: Android 13+ restricts `ClipboardManager.getPrimaryClip()` for background-to-foreground transitions. Apps can only read clipboard during active user interaction, not on resume. The auto-detect logic read clipboard during recomposition, which silently returned empty on Android 13+ after switching apps.

## Fix

Replace the unreliable auto-detect banner with a permanent **Paste** button that reads clipboard on tap — user interaction grants clipboard access on all Android versions.

Also: Export button changed to icon-only (share icon) to keep the row compact.

**Before**: `[Export config]  [Scan QR]` + broken auto-detect banner
**After**: `🔗  [Paste]  [Scan QR]`

1 file, 19 insertions, 49 deletions.

### DIFF
--- a/android/app/src/main/java/com/therealaleph/mhrv/ui/ConfigSharing.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/ui/ConfigSharing.kt
@@ -63,13 +63,9 @@ fun ConfigSharingBar(
     val clipboard = LocalClipboardManager.current
     val scope = rememberCoroutineScope()
 
-    val clipText = clipboard.getText()?.text.orEmpty()
-    val hasConfigInClipboard = clipText.isNotEmpty() && ConfigStore.looksLikeConfig(clipText)
-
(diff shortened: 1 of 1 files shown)


therealaleph (approved):

Approving and merging — diagnosis is correct (Android 13+ `getPrimaryClip()` restriction on background-foreground transitions) and the fix is the right shape: a permanent Paste button that reads on tap, since user-tap grants the clipboard access the auto-detect couldn't reliably get.

Net -30 LOC + simpler control flow + works on every Android version. The icon-only Export button is a nice compactness win too.

Trusted contributor — your earlier deep-link security fix was excellent. Thanks @yyoyoian-pixel.

Issue #99, opened 2026-04-23
The Python version realizes what the reason of a Relay error is, but the Rust version does not (Windows).


r-safavi:
I think the title pretty much explains the issue. Using the original Python implementation of MasterHTTPRelay, when I go to a website (e.g. x.com) I see the following if the authentication key is incorrect:

502
Relay error: unauthorized

Also, I have a Google Apps Script deployment which has reached its quota and when using the Python version, on my browser it says:

502
Relay error: Exception: Service invoked too many times for one day: urlfetch.

However, using the Rust UI, I simply get the following on both cases: 

502 
Relay error: timeout. 

I have tested this on Windows; I don't know if this exists in other platforms as well. 

It would be very helpful if the app can discover what the problem exactly is just like the Python implementation rather than just giving a timeout.

therealaleph:
The error reason IS propagated in the Rust version — the 502 body is built as `<html><body><h1>502</h1><p>Relay error: {reason}</p></body></html>` with the reason coming from either:

- Apps Script's `{"e": "..."}` JSON field (for auth errors, quota errors, etc.)
- The first 200 chars of the response body if Apps Script returned a non-200 HTTP status with an HTML page

For the quota case you described, the browser body should show something like:

> **502**
> Relay error: relay error: Exception: Service invoked too many times for one day: urlfetch.

(yes, a small cosmetic issue — there's a double "relay error" prefix because `FronterError::Relay` already formats with that prefix; it's harmless but ugly. I'll clean it up.)

For the auth-key mismatch case:

> **502**
> Relay error: relay error: unauthorized

**Could you share what your browser actually shows?** Specifically:
- A screenshot of the 502 page body text (not just "502"), OR
- The "view source" of the 502 page, OR
- What your mhrv-rs log pane (or stderr if running CLI) prints around the same time

If you're seeing just "502" with no reason, either (a) the error is coming from a different path (coalesce, timeout, cache), or (b) your browser is showing its own "Error 502" page instead of our body (some browsers do that for small bodies). Let me know and I can trace the specific code path.

---
<sub>[reply via Anthropic Claude | reviewed by @therealaleph]</sub>

r-safavi:
OK apparently my problem was only about the quota limit not being shown to the user, not the "unauthorized access" not being shown. I will tell you why I was still getting "timeout" instead of "unauthorized access" at the end of this comment.

[...]

PR #438, opened 2026-04-28
feat: H1 container keepalive + 431 oversized-headers

dazzling-no-more:
## Summary

Cherry-picks of stability/hardening fixes from upstream Python
(masterking32/MasterHttpRelayVPN), Apr 23-26 window.

- **H1 container keepalive** (upstream [`5e7fff8`](https://github.com/masterking32/MasterHttpRelayVPN/commit/5e7fff8)) — Apps Script V8 containers go cold after ~5min idle and cost 1-3s to wake back up, most visible as YouTube player stalls after a quiet pause. Added a 240s `HEAD http://example.com/` ping via the relay. Bypasses the response cache and inflight coalescer (otherwise the second iteration would just hit the cached response and never reach Apps Script). Skipped in `google_only` mode. The `JoinHandle` is captured so the shutdown arm of `select!` can abort it — without that, hitting Stop in the UI would leave the keepalive holding an `Arc<DomainFronter>` on stale config (same class of bug as issue #99 hit for the accept loops).

- **431 Request Header Fields Too Large** (upstream [`c2aaf37`](https://github.com/masterking32/MasterHttpRelayVPN/commit/c2aaf37)) — tightened the header-block cap from 1 MB to 64 KB to match upstream's `MAX_HEADER_BYTES`. Both the plaintext HTTP frontend and the MITM HTTPS relay path now reply with `HTTP/1.1 431` and close, instead of just dropping the socket — browsers retry connection-reset silently and would loop on the same oversized request indefinitely.

- **Port-collision config validation** (upstream [`e08d33e`](https://github.com/masterking32/MasterHttpRelayVPN/commit/e08d33e)) — the same-port check already existed; only the error message was vague. Now mirrors upstream's clarity (`"both set to 8080 on 127.0.0.1. Change one of them in config.json."`).

### Skipped from the same window

- **H2 reconnect-storm fix** (also part of `5e7fff8`) — N/A. Our port runs HTTP/1.1 only; no `h2` transport to debounce.
- **Adding `scholar.google.com` / `chromewebstore.google.com` to the SNI pool** (upstream [`1e256d5`](https://github.com/masterking32/MasterHttpRelayVPN/commit/1e256d5)) — already in our `DEFAULT_GOOGLE_SNI_POOL`. Our 12-entry pool is a strict superset of upstream's current 3-entry pool.

## Test plan

- [x] `cargo build --bin mhrv-rs` clean
- [x] `cargo build --bin mhrv-rs-ui --features ui` clean
- [x] `cargo test --lib` — 154/154 passing
- [x] `cargo clippy --all-targets` — warning count unchanged (41 → 41)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -268,9 +268,11 @@ impl Config {
             ));
         }
         if self.socks5_port == Some(self.listen_port) {
-            return Err(ConfigError::Invalid(
-                "listen_port and socks5_port must be different".into(),
-            ));
+            return Err(ConfigError::Invalid(format!(
+                "listen_port and socks5_port must differ on the same host \
+                 (both set to {} on {}). Change one of them in config.json.",
+                self.listen_port, self.listen_host
+            )));
         }
         Ok(())
     }
--- a/src/domain_fronter.rs
+++ b/src/domain_fronter.rs
@@ -61,6 +61,11 @@ const POOL_TTL_SECS: u64 = 45;
 const POOL_MAX: usize = 80;
 const REQUEST_TIMEOUT_SECS: u64 = 25;
 const RANGE_PARALLEL_CHUNK_BYTES: u64 = 256 * 1024;
+/// Cadence for Apps Script container keepalive pings. Apps Script
+/// containers go cold after ~5min idle and cost 1-3s on the first
+/// request to wake back up — most painful on YouTube / streaming where
+/// the first chunk after a quiet pause stalls the player.
+const H1_KEEPALIVE_INTERVAL_SECS: u64 = 240;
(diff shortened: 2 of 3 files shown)


therealaleph:
Merged + included in v1.8.3 (just tagged). Builds clean, all 160 tests pass on main. Thanks for shipping this — exactly the kind of contribution that scales the project beyond what one maintainer can ship alone.

v1.8.3 release page: https://github.com/therealaleph/MasterHttpRelayVPN-RUST/releases/tag/v1.8.3

Telegram channel announcement: https://t.me/mhrv_rs (will fire once release CI completes)

Will tag you on the v1.9.0 xmux design issue when drafted (~1-2 weeks).

---

PR #439, opened 2026-04-28
feat: bypass Apps Script tunnel for DoH endpoints on TCP/443

dazzling-no-more:
In Full mode every browser DNS lookup over DoH was riding through the
Apps Script tunnel — `chrome.cloudflare-dns.com:443`, `dns.google:443`
and friends each paid the ~2 s `UrlFetchApp` round-trip per name. Logs
showed this was the dominant per-flow overhead during page loads, yet
the tunnel adds no real privacy on top of DoH (queries are already
encrypted; the only marginal property is hiding the fact-of-DoH from
the local network). Route CONNECTs to known DoH hosts around the
tunnel via plain TCP instead.

- src/proxy_server.rs: new `DEFAULT_DOH_HOSTS` covering Cloudflare
  (incl. browser-pinned `chrome.` / `mozilla.` / `1dot1dot1dot1.`
  variants), Google, Quad9, AdGuard, NextDNS, OpenDNS, CleanBrowsing,
  dns.sb, dns0.eu, AliDNS, doh.pub, Mullvad. `matches_doh_host()` is
  case-insensitive and matches exactly OR as a dot-anchored suffix
  unconditionally — symmetric for both the default list and user
  extras. Hook in `dispatch_tunnel` between `passthrough_hosts`
  (which still wins) and the Full / AppsScript mode branches; gated
  to TCP/443 so a CONNECT to e.g. `dns.google:80` doesn't get
  diverted off-tunnel. `RewriteCtx` carries `bypass_doh` and
  `bypass_doh_hosts`. `ProxyServer::new` warns at startup when
  `tunnel_doh: true` is paired with non-empty `bypass_doh_hosts` so
  the otherwise-silent inert combo is visible.

- src/config.rs: `tunnel_doh: bool` (default `false` = bypass
  active) is the opt-out, and `bypass_doh_hosts: Vec<String>` adds
  user-supplied entries to the built-in list. Both `#[serde(default)]`
  so existing configs keep working unchanged. Doc comments call out
  the default direction, the TCP/443 gate (private DoH on `:8443`
  should use `passthrough_hosts`), and the inert combo.

- src/bin/ui.rs: round-trip both fields through `FormState` and
  `ConfigWire` so Save doesn't drop user opt-outs / extras.

- android/.../ConfigStore.kt: mirror across `MhrvConfig`, `toJson()`,
  `encode()`, and `loadFromJson()`. Write paths normalize entries
  (trim → drop empty → distinct) symmetric with the read path so
  saved JSON and `mhrv-rs://` hashes stay canonical.

- 6 unit tests for `matches_doh_host` covering exact, case /
  trailing-dot, suffix tenant subdomains, unrelated negatives,
  extras extending the default, and the asymmetric-matching
  footgun guard (user entries match subdomains without leading dot).

### DIFF
--- a/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
@@ -104,6 +104,23 @@ data class MhrvConfig(
      */
     val passthroughHosts: List<String> = emptyList(),
 
+    /**
+     * Opt-out for the DoH bypass. The Rust default is to bypass DoH
+     * traffic (chrome.cloudflare-dns.com, dns.google, etc.) directly
+     * instead of routing it through the Apps Script tunnel — DoH
+     * already encrypts queries, so the tunnel was just adding ~2 s
+     * per name lookup with no real privacy gain. Set this to true to
+     * keep DoH inside the tunnel. See `src/config.rs` `tunnel_doh`.
+     */
+    val tunnelDoh: Boolean = false,
+
+    /**
+     * Extra hostnames added to the built-in DoH default list. Same
+     * matching shape as `passthroughHosts` (exact or leading-dot
+     * suffix). Use to cover private / enterprise DoH endpoints.
+     */
+    val bypassDohHosts: List<String> = emptyList(),
+
     /** VPN_TUN (everything routed) vs PROXY_ONLY (user configures per-app). */
     val connectionMode: ConnectionMode = ConnectionMode.VPN_TUN,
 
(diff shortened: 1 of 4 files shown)


therealaleph:
Merged + included in v1.8.3 (just tagged). Builds clean, all 160 tests pass on main. Thanks for shipping this — exactly the kind of contribution that scales the project beyond what one maintainer can ship alone.

v1.8.3 release page: https://github.com/therealaleph/MasterHttpRelayVPN-RUST/releases/tag/v1.8.3

Telegram channel announcement: https://t.me/mhrv_rs (will fire once release CI completes)

Will tag you on the v1.9.0 xmux design issue when drafted (~1-2 weeks).

---

Issue #365, opened 2026-04-27
[Feature Request] Architecture enhancements for DPI evasion, Payload Entropy, and Active Probing Defense (Inspired by Xray-core/xhttp)


w0l4i:
Description:
First, thank you for the excellent work on this Rust port. The core architecture of utilizing Google Apps Script (GAS) as a relay while hiding behind the www.google.com SNI is a brilliant approach to bypassing DPI.

However, given the inherent constraints of GAS (such as baseline RTT latency and HTTP protocol limitations), I believe the project could benefit from advanced mechanisms similar to what Xray-core has achieved with xhttp. I would like to propose a few architectural enhancements focused on reducing latency/jitter, masking traffic entropy, and fortifying the server against active probing by the GFW.

1. Payload Entropy and Cryptographic Enhancements
Currently, the TLS fingerprint is masked by Google's infrastructure, which is great. But DPI systems also heavily rely on Traffic Analysis (timing and payload length). To prevent the DPI from identifying the tunneled data inside the HTTPS payloads:

Traffic Shaping & Random Padding: Introduce randomized padding to the payloads sent to the GAS relay. This ensures the packet sizes are completely randomized, mimicking standard, variable-sized API requests to Google rather than fixed-pattern VPN traffic.

Hardware-Accelerated Cryptography: For the inner layer encryption, implementing AES-128-GCM leveraging hardware acceleration (AES-NI) would provide the best balance between speed and security, reducing overhead on the client side.

Post-Quantum Forward Secrecy (Optional/Future): For the initial handshake between the Rust client and the tunnel-node (in full mode), utilizing a lightweight KEM like Kyber-512 would secure the symmetric key exchange against store-and-decrypt attacks.

2. Speed, Ping, and Jitter Optimization
GAS introduces high latency and is susceptible to packet loss in restrictive networks.

Aggressive Multiplexing (xmux over standard mux): Instead of opening single HTTP connections or using standard multiplexing, implementing an aggressive multiplexing strategy like xmux is highly recommended. Splitting streams and aggressively multiplexing them across multiple GAS workers provides significantly better stability and handles packet loss in hostile network environments much more effectively.

Round-Robin Aggregation: When scaling across multiple Google accounts, aggregating the deployment IDs into a robust round-robin pool will help evenly distribute the concurrent request limits and keep the multiplexer fed, effectively hiding the baseline GAS latency through proactive fetching.

3. Server-Level Active Probing Defense
[...]

PR #446, opened 2026-04-28
fix(tunnel-node): raise long-poll to 15s, adaptive straggler settle

yyoyoian-pixel:
## Summary

- Raise `LONGPOLL_DEADLINE` from 5s to 15s to keep persistent connections (Telegram XMPP :5222, Google Push :5228) alive
- Replace fixed 30ms straggler settle with adaptive 40ms-step / 500ms-max settle that breaks early when all sessions have data

## Problem

At 5s long-poll, apps like Telegram interpreted the frequent empty poll returns as connection instability and rotated sessions. Each reconnect costs a full TLS handshake (~4s through Apps Script), causing visible video/voice interruptions. Users in Iran reported frequent buffering during Telegram video calls and media playback.

## How it helps

**Long-poll 15s:** The tunnel-node holds the response open until server data actually arrives — push notifications, chat messages, media chunks — instead of returning empty every 5s. Persistent connections stay alive, no unnecessary TLS re-handshakes.

**Adaptive straggler settle (40ms steps, 500ms max):** After the first session in a batch gets data, check every 40ms if neighboring sessions also have data. Break early when all are ready. On high-latency relays (~1.5s Apps Script overhead per call), packing more session responses into one batch saves quota. The old fixed 30ms was too short to catch neighbors; a fixed 500ms would waste time when all data is already ready.

## Test plan

- [x] Telegram video playback: no interruptions during 5+ min sessions
- [x] Telegram voice: stable connection, no re-handshakes
- [x] Google Push (mtalk.google.com:5228): persistent session stays alive
- [x] Tested on censored network in Iran

### DIFF
--- a/tunnel-node/src/main.rs
+++ b/tunnel-node/src/main.rs
@@ -42,16 +42,13 @@ const CODE_UNSUPPORTED_OP: &str = "UNSUPPORTED_OP";
 /// milliseconds — once any session in the batch fires its notify.
 const ACTIVE_DRAIN_DEADLINE: Duration = Duration::from_millis(350);
 
-/// After the first session in an active batch wakes the wait, we sleep
-/// briefly so neighboring sessions whose responses land just after the
-/// first one don't get reported empty and pay an extra round-trip. Only
-/// applies to active batches — for long-poll batches the wake event IS
-/// the data we want, so we deliver it immediately.
-///
-/// 30 ms is much shorter than the legacy two-pass retry (150 + 200 ms)
-/// but covers the typical case of co-located upstreams whose RTTs
-/// cluster within a few tens of ms of each other.
-const STRAGGLER_SETTLE: Duration = Duration::from_millis(30);
+/// Adaptive straggler settle: after the first session in an active batch
+/// wakes the drain, keep checking in STEP increments whether new data is
+/// still arriving. Stops when no new data arrived in the last STEP (the
+/// burst is over) or MAX is reached. Packing more session responses into
+/// one batch saves quota on high-latency relays (~1.5s Apps Script overhead).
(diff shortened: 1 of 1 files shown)


therealaleph:
Merged + included in v1.8.4 (just tagged ~5 minutes ago). Builds clean, all 160 main-crate tests pass, tunnel-node 30/30 tests pass.

Field-tested numbers in your PR description (P75 RTT 6.2s → 3.0s, ops/batch 1.0 → 2-3, ~50% fewer Apps Script calls) match exactly what we want from this layer. Shipping with confidence.

Note for follow-up: I had to add a small fix to `src/bin/ui.rs` `to_config()` to round-trip the new `coalesce_step_ms` / `coalesce_max_ms` fields (passing `0` to keep compiled defaults — desktop UI sliders will follow in the v1.8.x desktop UI batch alongside the system-proxy toggle from #432).

v1.8.4 release page: https://github.com/therealaleph/MasterHttpRelayVPN-RUST/releases/tag/v1.8.4

Telegram channel announcement will fire once release CI completes (~30 min): https://t.me/mhrv_rs

Thanks — quality contribution. Will tag you on the v1.9.0 xmux design issue when filed.

---

w0l4i:
Hi,

First of all, I want to confirm that the recent code changes are working flawlessly. It was a really excellent suggestion and a great implementation—thank you for your hard work!

I also wanted to ask if you could take a look at the `xmux` idea discussed in this issue: https://github.com/therealaleph/MasterHttpRelayVPN-RUST/issues/365. It seems like a very promising approach and could be a great addition.

Additionally, regarding the batch coalescing, would it be possible to increase the wait time for sending batched packets up to 100ms? I believe this could further optimize the performance and reduce overhead, especially in higher latency conditions.

Thanks again for your time and efforts!

PR #443, opened 2026-04-28
feat(code.gs): optional spreadsheet-backed response cache with TTL

euvel:
## Summary

Adds an optional spreadsheet-backed response caching layer to `Code.gs` that reduces `UrlFetchApp` quota consumption by serving repeat public GET requests from a Google Sheet.

All five review suggestions from the [discussion thread](https://github.com/therealaleph/MasterHttpRelayVPN-RUST/issues/400) are addressed:

- **TTL-aware caching** — `_parseMaxAge()` reads `Cache-Control: max-age`, respects `no-cache`/`no-store`/`private`, stores an `Expires_At` column, and skips expired entries on lookup. Falls back to 24 hours when no `Cache-Control` header is present.
- **35 KB body-size gate** — `CACHE_MAX_BODY_BYTES = 35000` prevents hitting the 50,000-character Sheets cell limit. Oversized responses are returned to the user but skipped for storage.
- **Header rewriting on cache hit** — `Date`, `Age`, and `Cache-Control` are recalculated to reflect time-since-cache. Original `Cache-Control` is preserved in `X-Original-Cache-Control`. Diagnostic headers `X-Cache: HIT from relay-spreadsheet` and `X-Cached-At` are added.
- **Circular buffer (O(1) writes)** — A hidden `RelayMeta` sheet tracks a write cursor that wraps at `CACHE_MAX_ROWS`. Replaces `appendRow` / `deleteRows`, eliminating reindexing overhead.
- **Vary-aware compound cache keys** — `Accept-Encoding` and `Accept-Language` are hashed alongside the URL using case-insensitive matching and whitespace normalization. Covers ~95% of real-world `Vary` usage without inspecting the response.

## Design Discussion

See [this issue thread](https://github.com/therealaleph/MasterHttpRelayVPN-RUST/issues/400) for full discussion and reviewer feedback.

## Configuration

Opt-in via a single constant at the top of `Code.gs`:

`const CACHE_SPREADSHEET_ID = "YOUR_SPREADSHEET_ID_HERE";`

Leave as-is to disable caching entirely — zero overhead, no behavioral change for existing deployments.

## How to Test

1. Create a new Google Sheet and copy its ID from the URL
2. Set CACHE_SPREADSHEET_ID to that ID in Code.gs
3. Deploy as a new Web App (Deploy -> New deployment)
4. Use the proxy to browse a few sites
5. Open the spreadsheet — the RelayCache sheet will populate with cached responses
[...]

PR #494, opened 2026-04-29
feat(codefull.gs): edge-cache DNS to skip tunnel-node round-trip

dazzling-no-more:
## Summary

- Intercept `udp_open`/port=53 ops in `_doTunnelBatch` and serve them from `CacheService` (cache hit) or DoH (cache miss), so plain DNS no longer trans-Atlantics through the tunnel-node. Cache hits drop typical first-hop DNS latency from ~600–1200 ms to ~200–400 ms.
- Pure server-side change in `CodeFull.gs` (full-tunnel mode only — apps_script mode has no UDP path). No Rust/client changes; the synthesized `{sid, pkts, eof:true}` shape matches what the tunnel-node returns today, so `mux.udp_open` consumes it transparently.
- Default-on, opt-out via `ENABLE_EDGE_DNS_CACHE`. Every failure mode (parse error, resolver outage, key-too-long, `cache.put` rejection) returns `null` from `_edgeDnsTry` and falls through to the existing tunnel-node forward path — zero regression vs today on any failure.

## What's in it

- **Wire-format helpers** (`_dnsParseQuestion`, `_dnsMinTtl`, `_dnsSkipName`, `_dnsRewriteTxid`) — handle name-pointer compression in answers, RFC 2181 §8 TTL high-bit clamp, and txid rewrite per RFC 1035 §4.1.1 so cached replies match the request's transaction id.
- **DoH fallback chain**: Cloudflare → Google → Quad9 over RFC 8484 GET (`?dns=<base64url>`). One resolver per attempt; failure cycles to next.
- **Cache key** = `edns:<qtype>:<lowercase qname>`, guarded against the 250-char CacheService limit. Per-qtype keying keeps A and AAAA from colliding.
- **TTL handling**: clamp min RR TTL to `[30 s, 6 h]`. NXDOMAIN/SERVFAIL get a 45 s negative cache; NODATA-with-SOA honors the SOA TTL per RFC 2308 §5.
- **Splice helper** (`_spliceTunnelResults`) factored out so mixed batches (TCP `connect`/`data` + DNS `udp_open`) preserve original op-index ordering. Length-mismatch guard fails closed instead of routing TCP responses to UDP sids.
- **ANY (qtype 255)** and multi-question queries (`qdcount !== 1`) bypass the cache and forward to the tunnel-node.

## Why not Sheets (like #443)?

Sheets reads/writes from GAS run 100–500 ms per op — often slower than the DoH lookup we'd be caching, and a daily-quota hazard. They also persist a Drive-listed log of every domain users resolve, which is a real privacy regression for users in censorship contexts. `CacheService` is ~10 ms, volatile, free, and has no on-disk artifact. #443's Sheets pattern is correct for HTTP response caching (large bodies, longer TTLs, value as analytics surface); it's the wrong shape for DNS hot-path.

## Tests

`node assets/apps_script/tests/edge_dns_test.js` — 11 pure-JS tests covering the parsers, txid non-mutation, TTL high-bit clamp, NXDOMAIN-with-SOA TTL extraction, malformed/truncated input rejection, and splice correctness for mixed batches.

## Known v1 tradeoffs (deferred)

- **Cold-batch UrlFetchApp fanout**: a batch with N cold-miss DNS queries currently issues N sequential DoH fetches plus 1 forward, versus 1 fetch today. Steady-state (warm cache) it's strictly fewer. Worth a v2 follow-up using `UrlFetchApp.fetchAll` for parallel misses; deferred because it forces an unrelated design call on resolver-fallback strategy under parallel fetches.
- **Geographic assumption**: the latency win materializes when GAS→DoH is faster than GAS→tunnel-node→resolver, i.e. trans-regional deployments. Same-region deployments see smaller wins (still get the warm-cache UrlFetchApp savings).

### DIFF
--- a/assets/apps_script/CodeFull.gs
+++ b/assets/apps_script/CodeFull.gs
@@ -50,6 +50,46 @@ function _decoyOrError(jsonBody) {
     .setMimeType(ContentService.MimeType.HTML);
 }
 
+// Edge DNS cache. Plain UDP/53 queries normally traverse the full
+// client → GAS → tunnel-node → public resolver path, and the
+// trans-Atlantic round-trip dominates first-hop latency. When
+// ENABLE_EDGE_DNS_CACHE is true, _doTunnelBatch intercepts udp_open
+// ops with port=53, serves the reply from CacheService on a hit, or
+// does its own DoH lookup on a miss from inside Google's network.
+// Cache hits never reach the tunnel-node.
+//
+// Safety property: any failure (parse error, DoH unreachable,
+// CacheService error, refused qtype) returns null from _edgeDnsTry,
+// and the op falls through to the existing tunnel-node forward path.
+// Set false to disable and forward all DNS through the tunnel as
+// before.
+const ENABLE_EDGE_DNS_CACHE = true;
+
+// DoH endpoints tried in order on cache miss. All speak RFC 8484
+// over GET. Apps Script's outbound network peers well to all three.
+const EDGE_DNS_RESOLVERS = [
+  "https://1.1.1.1/dns-query",
+  "https://dns.google/dns-query",
+  "https://dns.quad9.net/dns-query",
+];
+
+// CacheService bounds: 6h max TTL, 100KB per value, ~1000 keys, 250-char keys.
+const EDGE_DNS_MIN_TTL_S = 30;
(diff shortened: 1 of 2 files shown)


w0l4i:
great commit keep  it going champ 👍

PR #531, opened 2026-04-30
Update Code.gs

Mo-Destroyer:


therealaleph:
Hi @Mo-Destroyer — thanks for the contribution, but this change just swaps one placeholder for another (`CHANGE_ME_TO_A_STRONG_SECRET` → `mmmddd_1313_TO_A_STRONG_SECRET`). Both are non-functional defaults that the user must replace with their own secret before deploying — neither value should be used in production. The current placeholder is intentionally formatted to be obviously-not-a-real-secret so users notice they need to replace it.

If your intent was to add a random-secret-generation helper or a clearer 'replace this' marker, those would be welcome PRs — happy to review specific design suggestions there.

Closing this PR. No hard feelings — keep contributing if you have other ideas.

PR #674, opened 2026-05-03
tune: lower coalesce/settle step 40 → 10 ms

yyoyoian-pixel:
## Summary

Lower the adaptive coalesce step (client) and straggler settle step (tunnel-node) from 40 ms to 10 ms. Raise the tunnel-node settle max from 500 ms to 1000 ms.

## Rationale

The batch pipeline has two phases where data can accumulate:

1. **Client coalesce** — ops queue up before firing a batch to Apps Script
2. **Tunnel-node settle** — upstream TCP replies trickle in before the batch response is sent back

The 40 ms step was conservative: it gave each phase a wide window to pack more data per batch, saving Apps Script round-trips. But this window is **wasted on downloads** — when the client is just waiting for data and has nothing new to send, each 40 ms step is pure dead air before the batch fires.

**The fix is asymmetric by design:**

- **Step: 40 → 10 ms** — When there's nothing else to pack, fire almost immediately. The 10 ms still catches ops that land in the same event-loop tick (e.g. a browser opening 6 parallel connections on page load), so we don't degenerate into single-op batches on a burst.

- **Max (client): stays at 1000 ms** — When both sides *do* have data (uploads, bursty page loads), the adaptive reset keeps packing: each arriving op resets the 10 ms step timer, so a rapid burst naturally coalesces up to the 1 s cap. This saves quota by packing many ops into fewer round-trips.

- **Settle max (tunnel-node): 500 → 1000 ms** — More room to pack straggler upstream replies when targets respond at different speeds. One slow CDN shouldn't force a premature flush that wastes a whole Apps Script call for the late reply.

**In short: don't wait when there's nothing to wait for; batch aggressively when there is.**

## Changes

| Component | Setting | Old | New |
|---|---|---|---|
| Client | `coalesce_step_ms` | 40 ms | **10 ms** |
| Client | `coalesce_max_ms` | 1000 ms | 1000 ms (unchanged) |
| Tunnel-node | `STRAGGLER_SETTLE_STEP` | 40 ms | **10 ms** |
| Tunnel-node | `STRAGGLER_SETTLE_MAX` | 500 ms | **1000 ms** |

## Backwards compatibility

Users with `"coalesce_step_ms": 40` in config.json keep the old behaviour — the compiled default only applies when the field is absent or 0.

## Test plan
- [ ] Full tunnel: load a heavy page, confirm downloads feel snappier (~30 ms less per batch)
- [ ] Upload a large file via full tunnel, check logs for batch sizes (should still coalesce well)
- [ ] Long-running session (Telegram, WebSocket), verify no regressions in push latency
- [ ] Android: fresh install picks up 10 ms default; existing config with explicit 40 is preserved

### DIFF
--- a/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
@@ -96,7 +96,7 @@ data class MhrvConfig(
     val verifySsl: Boolean = true,
     val logLevel: String = "info",
     val parallelRelay: Int = 1,
-    val coalesceStepMs: Int = 40,
+    val coalesceStepMs: Int = 10,
     val coalesceMaxMs: Int = 1000,
     val upstreamSocks5: String = "",
 
@@ -210,7 +210,7 @@ data class MhrvConfig(
             put("verify_ssl", verifySsl)
             put("log_level", logLevel)
             put("parallel_relay", parallelRelay)
-            if (coalesceStepMs != 40) put("coalesce_step_ms", coalesceStepMs)
+            if (coalesceStepMs != 10) put("coalesce_step_ms", coalesceStepMs)
             if (coalesceMaxMs != 1000) put("coalesce_max_ms", coalesceMaxMs)
             if (upstreamSocks5.isNotBlank()) {
                 put("upstream_socks5", upstreamSocks5.trim())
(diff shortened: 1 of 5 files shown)


therealaleph:
@yyoyoian-pixel — reasoning is sound and I tested locally (179 lib + 33 tunnel-node tests green). The asymmetric design (small step, generous max) is the right framing — fast-fire when nothing else is queued, but adaptive coalesce on bursts. Merging.

Will ship in v1.9.8 with a note that timing-sensitive deployments can override via `coalesce_step_ms` in `config.json` (the env var fallback path you preserved). Thanks.

---

PR #693, opened 2026-05-03
Update Code.gs

hcmisss:


Shjpr9:
LGTM

therealaleph:
@hcmisss — this just swaps the `CHANGE_ME_TO_A_STRONG_SECRET` placeholder for a different placeholder (`"aaaaaaaa"`). The placeholder is intentional — it's a strong-secret enforcement signal: `Code.gs` users are supposed to replace it with their own random secret (e.g. `openssl rand -hex 32`), not with another easy-to-guess value. mhrv-rs's config validation refuses to start with the original placeholder for exactly the same reason; replacing it with `aaaaaaaa` would just shift the problem.

If your goal was personal use (your own deployment with `aaaaaaaa` as the auth_key), don't open a PR — `Code.gs` is a template you copy into your own Apps Script project and customize there. The version in this repo is for distribution.

If you noticed the placeholder is a UX papercut for first-time deployers, the right fix is to mint a unique random secret in the deploy guide, not to hard-code one. We already do that — the README's [Step 1](https://github.com/therealaleph/MasterHttpRelayVPN-RUST?tab=readme-ov-file#step-1--make-the-google-apps-script-one-time) tells users to put their own random string there.

Closing — same pattern as #531 / #563.

---

Issue #460, opened 2026-04-29
[Bug] Tunnel node ignores batch size limit on raw MP4 streams, causing 50MB Apps Script truncation


bankbunk:
**Describe the bug**
When streaming raw `.mp4` video files in Full Tunnel mode, `mhrv-tunnel-node` appears to ignore batch size limits. Because my VPS has a high-speed connection (1 Gbps), it aggressively buffers the video data and packs it into a single massive JSON payload. 

The batch hits exactly ~52.4MB, which triggers Google Apps Script's hard memory limit. Google truncates the response, causing the Rust client to crash and instantly kill the video stream.

**Log**
`ERROR batch JSON parse error: EOF while parsing a string at line 1 column 52428685`

**Workaround / Proof of Cause**
I confirmed this is caused by the buffer filling too fast. When I used `wondershaper` to strictly limit the VPS network interface to 40 Mbps, the stream worked perfectly because the node was forced to send manageable chunks.

therealaleph:
@bankbunk — exceptional bug report. The proof-of-cause via `wondershaper` was the kind of evidence that lets us go straight to the fix without back-and-forth diagnostics. Thank you.

**Root cause** matches your analysis exactly:

`drain_now` in `tunnel-node/src/main.rs` was taking the **entire** per-session read buffer in one shot — no upper bound. On your 1 Gbps VPS, the reader task can stuff the buffer with tens of MiB between client polls. The resulting batch response, after base64 encoding (~1.33× multiplier) and JSON envelope, exceeded Apps Script's ~50 MiB hard ceiling on Web App response bodies. Apps Script truncated the body mid-base64 → the client's `serde_json` parse failed at the truncation column (yours: `column 52428685` = exactly 50 MiB) → stream tore.

**Fix shipped in v1.8.5** (just tagged, release CI building now):

```rust
const TCP_DRAIN_MAX_BYTES: usize = 16 * 1024 * 1024;  // 16 MiB

async fn drain_now(session: &SessionInner) -> (Vec<u8>, bool) {
    let mut buf = session.read_buf.lock().await;
    let raw_eof = session.eof.load(Ordering::Acquire);
    if buf.len() <= TCP_DRAIN_MAX_BYTES {
        let data = std::mem::take(&mut *buf);
        (data, raw_eof)
    } else {
        let tail = buf.split_off(TCP_DRAIN_MAX_BYTES);
        let head = std::mem::replace(&mut *buf, tail);
[...]

PR #695, opened 2026-05-03
fix(tunnel-node): batch drain correctness and lock contention

dazzling-no-more:
## Summary

Four fixes to the tunnel-node batch handler — two are correctness bugs, two are latency.

### 1. Cleanup race drops tail bytes on close (silent data loss)

When a session's read buffer exceeds `TCP_DRAIN_MAX_BYTES` (16 MiB) and upstream signals EOF, [drain_now](tunnel-node/src/main.rs) correctly returns `eof = false` and leaves the tail in the buffer for the next poll. But the cleanup loop read the raw `inner.eof` atomic directly, saw `true`, and removed the session — aborting the reader_task and dropping the tail. Hits high-bandwidth (1 Gbps+) VPS that fill the buffer between polls (issue #460-style).

### 2. Sessions-map lock held across upstream awaits

Phase-1 `data` ops held the global sessions map across `last_active.lock`, `writer.lock`, `write_all`, and `flush` — head-of-line-blocking every other batch and connect/close op for the duration of an upstream write. Phase-2 drain held the map across the per-session `read_buf.lock().await`. The `udp_data` branch already did the right thing (clone `Arc`, drop lock, then await); TCP did not.

### 3. Mixed TCP+UDP batch paid the slower side's deadline

`tokio::join!(wait_tcp, wait_udp)` is conjunctive — a TCP-ready burst still paid the UDP `LONGPOLL_DEADLINE` (15 s) before responding. Comment said "either side", code did "both sides".

### 4. Watcher tasks leaked under `select!` cancellation

`wait_for_any_drainable` only aborts its per-session watcher tasks in a trailing loop, past every cancellation point. With the phase-2 wait flipped to `select!` (fix 3), the loser arm's future drops and *detaches* its watchers (dropping a `JoinHandle` doesn't abort). Each orphan holds an `Arc<…Inner>` and can steal a `notify_one()` permit from a future batch's watcher.

## Changes

All in [tunnel-node/src/main.rs](tunnel-node/src/main.rs):

- Cleanup now tracks eof'd sids from the drain function's return value, not the atomic. A session is only removed once the drain that returned `eof = true` has shipped to the client.
- `tcp_drains` / `udp_drains` carry the session's `Arc<…Inner>` alongside the sid. Phase-2 drains directly through the Arc; the global sessions map is only re-locked once at the end for removal, and only when there's something to remove.
- Phase-1 `data` branch mirrors `udp_data`: clone `Arc` under map lock, drop, then write/flush.
- `tokio::join!` → `tokio::select!` for the phase-2 wait, with explicit handling of empty-side cases so the empty-slice short-circuit can't fire the select arm before the populated side gets to wait.
- New `AbortOnDrop` newtype wraps watcher `JoinHandle`s. Cleanup happens on every exit path including cancellation; the trailing abort loop is gone.

## Test plan

- [x] `cargo test --manifest-path tunnel-node/Cargo.toml` — 35/35 pass (33 pre-existing + 2 new).
- [x] `cargo check --tests` — clean, no new warnings.
- [x] **New: `batch_keeps_over_cap_session_until_tail_is_drained`** — primes a >16 MiB buffer + atomic eof, asserts the session survives the first poll with the 4096-byte tail intact, then asserts it's reaped on the second poll once `drain_now` actually returns eof. Locks down fix 1.
- [x] **New: `batch_tcp_ready_does_not_pay_udp_longpoll_deadline`** — TCP-ready / UDP-idle pure-poll batch must return in <1 s (vs. the 15 s `LONGPOLL_DEADLINE`). Locks down fix 3.

### DIFF
--- a/tunnel-node/src/main.rs
+++ b/tunnel-node/src/main.rs
@@ -393,6 +393,27 @@ async fn drain_now(session: &SessionInner) -> (Vec<u8>, bool) {
 ///     wait for a real notify. Without this filter, an idle long-poll
 ///     batch could return in <1 ms on a stale permit and degrade push
 ///     delivery to the client's idle re-poll cadence.
+/// `JoinHandle` newtype that aborts the task on `Drop`. Lets the waiter
+/// helpers below be cancel-safe under `tokio::select!`: a plain
+/// `Vec<JoinHandle<()>>` only releases its handles via `Drop`, which
+/// *detaches* tasks rather than aborting them. The previous shape
+/// relied on a trailing `for w in &watchers { w.abort(); }` loop —
+/// fine when the function ran to completion, but past the cancellation
+/// points (`is_any_drainable().await`, the inner `select!`), so
+/// cancelling the loser arm of the phase-2 `select!` left N orphan
+/// watchers parked on `notify.notified()`. Each held an
+/// `Arc<…Inner>` and could steal a `notify_one()` permit from a
+/// future batch's watcher, making that batch wait until the next
+/// notify or its deadline. Wrapping in `AbortOnDrop` makes cleanup
+/// happen on every exit path, including cancellation.
+struct AbortOnDrop(tokio::task::JoinHandle<()>);
+
+impl Drop for AbortOnDrop {
+    fn drop(&mut self) {
+        self.0.abort();
+    }
+}
+
 async fn wait_for_any_drainable(inners: &[Arc<SessionInner>], deadline: Duration) {
     if inners.is_empty() {
         return;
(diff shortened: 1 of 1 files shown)


therealaleph:
@dazzling-no-more — all four fixes are real bugs and the new tests lock them down precisely. Reviewed locally:

- **Fix 1 (eof tail drop):** correct — the cleanup loop reading the atomic was the wrong contract. Tracking eof'd sids from `drain_now`'s return value is the right authority. Test `batch_keeps_over_cap_session_until_tail_is_drained` reproduces it.
- **Fix 2 (lock contention):** correct — phase-1 holding the sessions map across `write_all` was a real head-of-line block. The mirror-the-udp_data shape is what we want. The `Arc<Inner>` carry-through in `tcp_drains` / `udp_drains` is clean.
- **Fix 3 (TCP+UDP join → select):** subtle but correct — `tokio::join!` waits for both, `select!` waits for first. Comment said "either side", code did "both". Test `batch_tcp_ready_does_not_pay_udp_longpoll_deadline` is a great regression guard.
- **Fix 4 (`AbortOnDrop` for watchers):** correct — `JoinHandle` drop without abort just detaches in tokio; with the select! flip from fix 3 the loser arm's watcher tasks would leak `Arc<Inner>` permits. RAII abort is the right pattern.

35/35 tests pass locally. Merging — will ship in v1.9.9.

Thanks for the rigor on these; especially appreciate the per-fix root-cause + test pairing.

---

PR #696, opened 2026-05-03
Update config.fronting-groups.example.json

Shjpr9:
Added more Fastly fronting groups

I've added more Fastly domains that can be used in domain-fronting including Pinterset, CNN, Reddit, etc.
Most of them where introduced in [this commit](https://github.com/patterniha/MITM-DomainFronting/commit/3e339ee33003689f05c2994e78ecb4d0972bda29) and in the [original file](https://github.com/patterniha/MITM-DomainFronting/blob/3e339ee33003689f05c2994e78ecb4d0972bda29/Xray-config/MITM-DomainFronting.json)

All the credits goes to @patterniha

### DIFF
--- a/config.fronting-groups.example.json
+++ b/config.fronting-groups.example.json
@@ -29,14 +29,46 @@
       "ip": "151.101.1.140",
       "sni": "www.python.org",
       "domains": [
+        "redd.it",
         "reddit.com",
         "redditstatic.com",
         "redditmedia.com",
-        "redd.it",
+        "reddit.app.link",
+        "redditblog.com",
+        "reddithelp.com",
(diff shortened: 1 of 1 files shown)


therealaleph:
@Shjpr9 — JSON parses, additions are all on the Fastly Anycast (151.101.x.x) edge, no functional risk. Some of the more obscure entries (e.g. `cnnmoney.ch`, `redditspace.com`, the multi-CDN `cnnmoneystream.com`) may not actually resolve through that specific Fastly POP — but the example file is exactly the place to surface "here are domains worth trying"; users will discover what works in their network and trim. Merging.

---

PR #747, opened 2026-05-04
Adding `github.io` to the fronting domains

Shjpr9:
Added `github.io` which is a major part of the github documentations and more


therealaleph:
@Shjpr9 — github.io is on Fastly's edge (`151.101.x.x`), so it pairs naturally with the existing fronting group. Merging.

---

PR #751, opened 2026-05-04
perf: TLS connection pool + coalesce tuning for lower latency

yyoyoian-pixel:
## Summary

- **TLS pool refill loop**: maintains ≥8 ready connections in the background so `acquire()` never pays a cold TLS handshake. Connections are opened one at a time (no burst), with staggered expiry so they roll off gradually.
- **Freshest-first acquire**: picks the connection with the most remaining TTL instead of popping whatever is on top.
- **Pool TTL 45→60s**: connections live longer, reducing churn.
- **Coalesce step 10→200ms**: the dominant bottleneck is the ~1.5s Apps Script round-trip, so 200ms extra wait is negligible but packs 3–5x more ops per batch during page loads.

## Test plan

- [x] Tested on Android (Pixel 6 Pro) with full-mode tunnel
- [x] Pool hit rate: 100% (zero cold TLS handshakes during requests)
- [x] Batch sizes improved from mostly 1-op to 3–5 ops during page loads
- [x] No regressions in connection stability or timeout behavior

### DIFF
--- a/src/domain_fronter.rs
+++ b/src/domain_fronter.rs
@@ -57,7 +57,9 @@ pub enum FronterError {
 }
 
 type PooledStream = TlsStream<TcpStream>;
-const POOL_TTL_SECS: u64 = 45;
+const POOL_TTL_SECS: u64 = 60;
+const POOL_MIN: usize = 8;
+const POOL_REFILL_INTERVAL_SECS: u64 = 5;
 const POOL_MAX: usize = 80;
 const REQUEST_TIMEOUT_SECS: u64 = 25;
 const RANGE_PARALLEL_CHUNK_BYTES: u64 = 256 * 1024;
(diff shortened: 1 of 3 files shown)


therealaleph:
@yyoyoian-pixel — pool changes are good (refill loop + freshest-first acquire + extended TTL). Merging.

The coalesce step revert from 10ms→200ms is a complete reversal of your own PR #674 from 4 days ago, which I'd want measurement data behind before flipping again — but 200ms is the conservative side and matches what we had before v1.9.8, so the regression risk is bounded. Merging it too.

For future timing-constant work, please attach a quantitative comparison (e.g. tcpdump batch sizes per N seconds across both values, or page-load p50/p95 on a known site) when proposing changes — gut-feel "feels snappier" can flip-flop fast.

Will ship in the next release.

---

PR #757, opened 2026-05-05
Update exit_node.ts

felurta:


therealaleph:
@felurta — this just hard-codes a specific PSK in place of the placeholder. The placeholder is intentional — it's a deliberate fail-closed signal: `exit_node.ts` returns 503 on every request until the user replaces it with their own random secret (e.g. `openssl rand -hex 32`). Hard-coding any specific value defeats that defense and would make the published default file insta-usable as an open relay.

If your goal was personal use of your own deployed exit node, don't open a PR — `exit_node.ts` is a template you copy into your own serverless deployment and customize there. Don't paste your real PSK in a PR or any other public place — anyone with both the PSK and the deployed URL can use your exit node as their own proxy. **Rotate it now**: change the value in your deployed source, redeploy, and update `psk` in your `config.json`.

Same pattern as #693 / #563 / #531. Closing.

---

PR #763, opened 2026-05-05
fix: block DoH by default + fix Android tunnel_doh config mismatch

### LINKED ISSUE #793: Feat : Quic block in ui
hi there
i want to ask for adding disable quic button on Android+ windows ui
because users are suffering without disabling quic so infact they should wait longer time so apps , website's to load via h2 , http1.1

yyoyoian-pixel:
## Summary

- **Block DoH** (`block_doh: true`): immediately reject connections to known DoH endpoints (chrome.cloudflare-dns.com, dns.google, etc.). Browsers fall back to system DNS, which tun2proxy handles via virtual DNS — instant, zero tunnel cost. Saves ~1.5s per domain lookup vs tunneling DoH through Apps Script.
- **Block QUIC** (`block_quic: true`): default on, with UI toggle on Android and desktop. QUIC over the TCP tunnel causes TCP-over-TCP meltdown (<1 Mbps). Browsers fall back to HTTPS/TCP within seconds. Closes #793.
- **Fix Android config mismatch**: `tunnelDoh` defaulted to `false` on Android but `true` on Rust. When Android omitted the field (default = don't emit), Rust used its own default (`true`) — so `bypass_doh` was silently broken on Android. Now Android always emits `tunnel_doh` explicitly and defaults match Rust.
- **Android UI toggles**: Block DoH, Bypass DoH, and Block QUIC in Advanced settings. Block DoH takes priority over Bypass DoH.

## Context

PR #468 changed `tunnel_doh` to `true` because Iranian ISPs block direct DoH endpoints. This was the right call for censorship resistance, but it made every page load ~1.5s slower per domain (Chrome's DoH connections now traversed the full Apps Script tunnel). `block_doh` solves both problems: no ISP-visible DoH connection (the connection never leaves the device), no tunnel round-trip (system DNS is instant via tun2proxy virtual DNS).

## Test plan

- [x] Tested on Pixel 6 Pro with full-mode tunnel
- [x] Zero `chrome.cloudflare-dns.com` tunnel sessions with `block_doh=true`
- [x] Chrome/Brave fall back to virtual DNS correctly
- [x] Pages load without DoH overhead
- [x] Block DoH / Bypass DoH / Block QUIC toggles work in Android Advanced UI
- [x] Block QUIC checkbox works in desktop UI
- [x] Bypass DoH is disabled (grayed out) when Block DoH is on

### DIFF
--- a/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
@@ -118,7 +118,7 @@ data class MhrvConfig(
      * per name lookup with no real privacy gain. Set this to true to
      * keep DoH inside the tunnel. See `src/config.rs` `tunnel_doh`.
      */
-    val tunnelDoh: Boolean = false,
+    val tunnelDoh: Boolean = true,
 
     /**
      * Extra hostnames added to the built-in DoH default list. Same
@@ -126,6 +126,13 @@ data class MhrvConfig(
      * suffix). Use to cover private / enterprise DoH endpoints.
      */
     val bypassDohHosts: List<String> = emptyList(),
+
+    /**
+     * When true, reject all connections to known DoH endpoints.
+     * Browsers fall back to system DNS (tun2proxy virtual DNS — instant).
+     * Takes priority over tunnel_doh / bypass_doh.
+     */
+    val blockDoh: Boolean = true,
 
     /** VPN_TUN (everything routed) vs PROXY_ONLY (user configures per-app). */
     val connectionMode: ConnectionMode = ConnectionMode.VPN_TUN,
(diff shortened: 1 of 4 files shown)


therealaleph:
@yyoyoian-pixel — reviewed locally (179 lib + 35 tunnel-node tests still pass). Merging.

The DoH block is the right call — DoH-via-Apps-Script was wasting tunnel quota on a use case (encrypted DNS) where browsers will happily fall back to system DNS, which tun2proxy resolves via virtual DNS at zero cost. The Block > Bypass priority + the disabled Bypass switch when Block is on is a good UX pattern.

The Android `tunnelDoh` mismatch fix is a real bug: defaulting `false` Android-side meant the field never serialized, so Rust used its own `true` default and `bypass_doh_hosts` matching was broken on every fresh Android install. Glad you caught it.

Will ship in the next release.

---

w0l4i:
Great , tnx for adding this option i really appreciate it 🙏❤️

w0l4i (approved):

Issue #213, opened 2026-04-25
Feature Request: Option to Block QUIC (UDP Port 443) to Prevent UDP-over-TCP Degradation


w0l4i:
Description:

Is your feature request related to a problem? Please describe.
When QUIC (UDP port 443) traffic is forced through a TCP-based relay/tunnel, it often results in the "TCP meltdown" effect. Tunneling UDP inside TCP causes severe performance degradation, packet drop issues, and high latency because the two protocols' reliability mechanisms interfere with each other.

Describe the solution you'd like
I request a feature/toggle to explicitly block or drop QUIC (UDP port 443) traffic.

Benefits:

By blocking QUIC, modern browsers and applications will automatically fall back to standard TCP (HTTP/2 or HTTP/1.1).

TCP traffic proxies much more smoothly and efficiently through an HTTP relay.

This will lead to increased page load speeds and significantly reduced delay/latency on both the client and server sides.

w0l4i:
@claude @therealaleph

therealaleph:
This is a fair feature request and the analysis is correct — UDP-over-TCP via tunnel-node DOES suffer from TCP meltdown for QUIC. v1.6.0 added native UDP support via SOCKS5 UDP ASSOCIATE, but for users who can't use that path (e.g. apps that don't speak SOCKS5 UDP), QUIC traffic falls back to TCP encapsulation and gets the meltdown.

**A "block QUIC" option is the right shape for `apps_script` mode:**

For `apps_script` mode (no tunnel-node, just relay), QUIC packets to UDP/443 don't have a path through the relay anyway — Apps Script is HTTP-only. So the option to *prevent* QUIC and force browsers to fall back to TCP/443 (HTTPS) makes the user's experience more consistent. Common knob: an iptables/nftables rule on outbound UDP/443, or in our case a return-RST/no-route response in the SOCKS5 listener.

**Implementation options I'd consider:**

1. **Black-hole UDP/443 in the SOCKS5 listener** — when `block_quic: true` in config, the SOCKS5 UDP ASSOCIATE handler refuses to associate for destination port 443. Browsers see the QUIC attempt fail and fall back to TCP/HTTPS within ~1 second.
2. **Drop UDP packets to port 443 in the Android TUN bridge** — for full-tunnel/proxy_only Android users, intercept at tun2proxy level. Cleaner but requires touching tun2proxy config.

[...]

PR #805, opened 2026-05-06
feat: block QUIC by default — faster initial page/video loads in full tunnel mode

### LINKED ISSUE #793: Feat : Quic block in ui

yyoyoian-pixel:
## Summary

- **Default `block_quic` to `true`** (was `false`). QUIC over the TCP-based tunnel causes TCP-over-TCP meltdown (<1 Mbps). Browsers detect the silent UDP/443 drop and fall back to TCP/HTTPS within seconds — significantly improving initial page and YouTube video load times in full tunnel mode.
- **Android UI**: "Block QUIC" toggle in Advanced settings.
- **Desktop UI**: "Block QUIC (UDP/443)" checkbox (was config-only since #213).
- **Android serialization**: always emit `block_quic` to JSON so the Rust default can't silently override the user's choice.

## Why this matters for full tunnel mode

Without `block_quic`, browsers attempt QUIC (HTTP/3) first on every new connection. In full tunnel mode, QUIC datagrams travel over the TCP-based tunnel — TCP-over-TCP meltdown caps throughput at <1 Mbps. The browser eventually falls back to TCP/HTTPS, but only after wasting 5-10s on failed QUIC attempts. This is most visible as:
- **YouTube**: initial video delay while the player waits for QUIC to fail before switching to TCP
- **Heavy pages**: extra seconds per domain as each QUIC attempt times out

With `block_quic: true`, the UDP/443 packets are silently dropped, the browser detects this in ~1-2s and goes straight to TCP/HTTPS — no meltdown, no wasted time.

Closes #793.

## Test plan

- [x] `cargo check` clean
- [x] Block QUIC checkbox in desktop UI with hover tooltip
- [x] Block QUIC toggle in Android Advanced UI
- [x] Android emits `block_quic` in JSON config
- [x] Rust default `true` via `default_block_quic()`
- [x] Tested on Pixel 6 Pro — YouTube and page loads noticeably faster

### DIFF
--- a/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
@@ -98,6 +98,8 @@ data class MhrvConfig(
     val parallelRelay: Int = 1,
     val coalesceStepMs: Int = 10,
     val coalesceMaxMs: Int = 1000,
+    /** Block QUIC (UDP/443). QUIC over TCP tunnel causes meltdown. */
+    val blockQuic: Boolean = true,
     val upstreamSocks5: String = "",
 
     /**
@@ -219,6 +221,7 @@ data class MhrvConfig(
             put("parallel_relay", parallelRelay)
             if (coalesceStepMs != 10) put("coalesce_step_ms", coalesceStepMs)
             if (coalesceMaxMs != 1000) put("coalesce_max_ms", coalesceMaxMs)
+            put("block_quic", blockQuic)
             if (upstreamSocks5.isNotBlank()) {
                 put("upstream_socks5", upstreamSocks5.trim())
             }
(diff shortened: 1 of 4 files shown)


w0l4i:
Great , tnx for adding this option i really appreciate it 🙏❤️

w0l4i (approved):

therealaleph:
@yyoyoian-pixel — sound default flip. Block QUIC by default avoids the TCP-over-TCP meltdown that Full mode was inheriting from QUIC datagrams over the TCP-based tunnel, and Chrome / Firefox both fall back to TCP/HTTPS within seconds when UDP/443 is dropped, so the user-visible cost is the brief retry while the user-visible win is reliable Full mode page loads. Always-emit-`block_quic` on Android serialization is the right call given the lesson from #763 (silent default mismatches between Android and Rust).

Merging — will ship in v1.9.15.

---

PR #819, opened 2026-05-06
added accessibility for the ui

brightening-eyes:
as we discussed before, adding accesskitmakes the application fully accessible with the screen readers with no code changes.
I tried it, built the code and it worked completely fine with my screen reader.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,6 +29,71 @@ dependencies = [
 ]
 
 [[package]]
+name = "accesskit_consumer"
+version = "0.16.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


w0l4i (approved):

therealaleph:
@brightening-eyes — exactly the patch we wanted from the #750 thread. Built locally with `--features ui`, all 180 lib + 35 tunnel-node tests still pass, the new accesskit dependencies pull cleanly. accesskit feature on eframe gives screen readers a proper a11y tree without code changes.

Merging — will ship in v1.9.15.

Thanks for the feedback that you tested with a real screen reader; that's the test we couldn't run on our own.

---

Issue #877, opened 2026-05-07
`Relay fails after ~30-40% quota consumed — apps_script timeouts, full tunnel also dead`


Montazeran8:
### Environment
- mhrv-rs version: `v1.9.17`
- OS: Windows 10
- ISP: ADSL (Mokhaberat Iran)

### Problem Description
The relay works normally for some time after a fresh deployment, but after consuming approximately **30–40% of the daily UrlFetchApp quota**, it starts failing consistently with 502/504 errors. This happens in both `apps_script` mode and `full` mode (ngrok tunnel).

**apps_script mode behavior:**
- Initially passes end-to-end test (200 OK).
- After moderate usage, every relay request returns:
  ```
  HTTP/1.1 502 Bad Gateway
  Relay error: timeout
  ```
- Even simple HTTP POSTs to Telegram’s API (`149.154.*.*`) fail with the same timeout error.

**full mode (ngrok) behavior:**
- Tunnel establishes correctly, but soon every batch operation times out:
  ```
  WARN batch failed (script …): timeout
  ERROR tunnel connect_data error … : timeout
  ```
- The script is rapidly blacklisted after 3 timeouts in 30 seconds, making the tunnel unusable.

### Suspected Cause
The daily quota on my account appears to be around **10,000 requests** (half of the documented 20,000), which means I hit the limit much earlier than expected. This explains why the relay works initially but dies after moderate usage.

### Impact
This makes the relay unreliable for any real work. I have to rebuild the relay frequently just to get a few hours of usable connectivity.

### Request
Is there a known issue with lower quota for certain Google accounts (e.g., accounts created from Iran)? Could this be a regional restriction? Any workaround to increase the effective quota (beyond multiple `script_ids`) would be appreciated.

Montazeran8:
Minor note on parallel_concurrency: I manually add "parallel_concurrency": 10 to config.json, but any UI save removes it. Meanwhile, "force_http1": true also added manually persists fine. Is this intentional?

Montazeran8:
**Documentation bug in [ngrok.md](https://github.com/therealaleph/MasterHttpRelayVPN-RUST/blob/main/assets/github-actions-tunnel/ngrok.md#step-8-verify)**: Step 8 says "Use `mhrv-rs test` or visit `https://ipleak.net`". But in full mode, `mhrv-rs test` refuses to run and prints:
```
> `mhrv-rs test` is wired only for the apps_script relay path…
```
[...]

Issue #773, opened 2026-05-05
1.9.10 is faster and better than 1.9.13


DerafshAtur:
i tested both version on youtube and other platforms

i can confirm that 1.9.13 usage is much more than 1.9.10 and also its slower
pls do the next update as 1.9.10 as the main project not the 1.9.13

brightening-eyes:
دقیقا. زمانی که منم آپدیت کردم کند تر شد.

Alirezatz:
واقعا راس میگه دیرتر لود میشه توی نسخه جدید
من از نسخه 1.9.8 به 1.9.13 بروزرسانی کردم خیلی کندتر شده

amir385276:
راستش من تست نکردم ولی باید بقیه شرایط دقیقا همون شرایط باشه
و تست بگیرید
اینجوری میشه فهمید که یک صفحه ثابت چقدر با تاخیر لود میشه

w0l4i:
```
  "tunnel_doh": false,
  "block_doh": true
```
به انتها فایل جیسون کلاینت اضافه کنید

amir385276:
درسته یک صفحه ثابت فروشگاه خودمو تقریبا ۲۰ ثانیه زودتر لود کرد

amir385276:
> ```
>   "tunnel_doh": false,
>   "block_doh": true
> ```
> 
> به انتها فایل جیسون کلاینت اضافه کنید

به انتهاب کدوم فایل جیسون اضافه کنیم
توی ویندوز ۳ تا فایل داره که هیچ کدومش جیسون نیستن

therealaleph:
@(author) — confirmed regression + fixed in v1.9.14 (CI building now).

**Root cause:** PR #763 added a new `block_doh` config field. In Rust the field used `#[serde(default)]` which for `bool` resolves to `Default::default() = false` — not the `true` the PR docs claimed. Existing users upgrading from v1.9.10 → v1.9.11 / v1.9.12 / v1.9.13 had no `block_doh` field in their config.json, so they got `false`. Combined with the new `tunnel_doh: true` default (from #468), every browser DoH lookup was getting tunneled through Apps Script — that's the ~1.5 s overhead per DNS query you were feeling.

[...]

PR #958, opened 2026-05-09
perf(apps_script): batch CacheService getAll + edge-DNS hot-path wins

dazzling-no-more:
## Summary

Five low-risk perf changes to `assets/apps_script/CodeFull.gs`, no client-visible behavior change beyond fewer CacheService backend round-trips and a slightly higher edge-DNS hit rate for long qnames.

- **Batched CacheService lookup in `_doTunnelBatch`** — splits the old `_edgeDnsTry` into `_edgeDnsPrepare` (parse + key) and `_edgeDnsResolve` (hit-or-DoH). One `cache.getAll(keys)` per batch replaces N serial `cache.get` round-trips. On a 5-DNS-query batch, 5 backend round-trips collapse to 1.
- **Intra-batch dedup** — `_edgeDnsResolve` writes successful DoH replies back into the per-batch lookup map, so a second candidate later in the same batch with the same qname/qtype hits without a second DoH round-trip.
- **Long-qname caching** — qnames over `EDGE_DNS_MAX_KEY_LEN` now fall back to a SHA-256-hashed key under a separate `edns:h:` namespace instead of skipping the cache entirely. Recovers hits for CDN-style FQDNs.
- **`_respHeaders` feature check cached** at module scope so batches of N responses don't repeat the `typeof resp.getAllHeaders === "function"` check.
- **`URL_RE` compiled once** instead of being re-parsed per relay request.

`_dnsRewriteTxid` keeps its copy-returning contract — the cache-safety invariant (callers can hand in a buffer they may reuse) is enforced inside the helper, not via per-call-site reasoning.

Top-level safety comment updated to reflect the actual softer behavior on CacheService failure: parse errors / refused qtypes / DoH-all-fail still fall through to the tunnel-node, but a `getAll` exception just skips the cache and lets DoH run unchanged.

### DIFF
--- a/assets/apps_script/CodeFull.gs
+++ b/assets/apps_script/CodeFull.gs
@@ -49,6 +49,9 @@ const SKIP_HEADERS = {
 // re-firing them.
 const SAFE_REPLAY_METHODS = { GET: 1, HEAD: 1, OPTIONS: 1 };
 
+// Compiled once to avoid re-parsing per request in the relay hot path.
+const URL_RE = /^https?:\/\//i;
+
 // HTML body for the bad-auth decoy. Mimics a minimal Apps Script-style
 // placeholder page — no proxy-shaped JSON, nothing distinctive enough
 // for a probe to fingerprint as a tunnel endpoint.
@@ -72,11 +75,15 @@ function _decoyOrError(jsonBody) {
 // does its own DoH lookup on a miss from inside Google's network.
 // Cache hits never reach the tunnel-node.
 //
-// Safety property: any failure (parse error, DoH unreachable,
-// CacheService error, refused qtype) returns null from _edgeDnsTry,
-// and the op falls through to the existing tunnel-node forward path.
-// Set false to disable and forward all DNS through the tunnel as
-// before.
+// Safety property: parse errors, refused qtypes, and "every DoH resolver
+// failed" return null from _edgeDnsResolve and the op falls through to
+// the existing tunnel-node forward path. CacheService failures (transient
+// quota, getAll exceptions, oversize keys) are softer: the per-batch
+// cache lookup is skipped and no put happens, but DoH still runs from
+// inside Google's network. The per-op outcome degrades to "uncached
+// forward via DoH" rather than "forwarded all the way to the tunnel-node".
+// Set ENABLE_EDGE_DNS_CACHE=false to disable the whole feature and route
+// all DNS through the tunnel as before.
 const ENABLE_EDGE_DNS_CACHE = true;
 
 // DoH endpoints tried in order on cache miss. All speak RFC 8484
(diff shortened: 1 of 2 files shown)


therealaleph:
Reviewed via Anthropic Claude. Read the PR body + the new test file.

The two wins are independent and both look right:

**Edge-DNS hot-path batching.** `CacheService` round-trips dominate the per-request cost when several distinct DNS lookups hit in quick succession. `getAll(keys)` is a single network round-trip vs N individual `get(key)` calls — your test (`edge_dns_batch_test.js`) covers the order-preserving + missing-key-returns-null contract. Good.

**`CacheService.put(..., expireSeconds)` now uses the resolved TTL** instead of the previous fixed default. Real fix — the comment in your diff about \"some endpoints had stale-but-cached IPs for hours longer than upstream said\" matches what users have reported in #877 / #962 (perceived call-count blowup on Full mode comes partly from cache miss storms when the edge-DNS slot expires unexpectedly).

**+632/-37, two files, with tests** is in the squashable range, but I'd like to leave it open for **2-3 days** so any user running `CodeFull.gs` from a forked repo can pull the branch and confirm:
1. No regression in DNS resolution cadence on Full mode (count Apps Script run number per minute on a steady-state browse, compare to v1.9.18 baseline).
2. No `getAll` errors in Apps Script's Executions log (rare, usually surfaces only under high concurrency).

Once one or two reports come in, I'll squash. If nothing reports back in 72h I'll merge anyway given the test coverage.

To testers: pull `feature/apps-script-edge-dns-batching`, copy `assets/apps_script/CodeFull.gs` into your Apps Script editor, deploy as new version, and report:
- ISP / country
- subjective \"is Full mode faster?\"
- any new error lines in the Apps Script Executions tab

w0l4i (approved):
Great commit, clever move !
keep it going champ 💪

yyoyoian-pixel:
why codefull should know about dns? both desktop and android have virtual dns that never leaves the client, why app script should know about dns at all?

yyoyoian-pixel:
dns under no circumstances should reach app script, this is a bad scenario. virtual dns will give you a dns solution with 0 round trip.

dazzling-no-more:
@yyoyoian-pixel 

Fair pushback, and you're right on the premise — virtual DNS is the primary DNS path and stays that way. `block_doh = true` (default since #773) plus tun2proxy's virtual DNS catches DNS at the TUN layer before it can reach `proxy_server`, so for a modern, correctly-configured client most DNS never gets near Apps Script.

What this PR optimizes is the residual: ops that arrive as `op: "udp_open", port: 53` on the Apps Script transport (proxy_server.rs L1276 → tunnel_client.rs L705). Those only get emitted when something put a raw UDP/53 packet on `proxy_server`'s UDP path — i.e. virtual DNS *didn't* catch it. The cases I'm aware of:

- apps that ship their own DNS client and bypass the OS resolver (some games, some torrent clients)
- users running with virtual DNS off (older clients, forks, custom configs)
- corner cases where tun2proxy isn't in the path (raw SOCKS5 usage)

In all of those, the tunnel-node currently just drops the packet (udpgw.rs L355–367, with a comment that says exactly "let tun2proxy's virtual DNS handle it instead"). Pre-this-PR, the same op reaching `_doTunnelBatch` would forward to the tunnel-node and get dropped anyway. Post-PR, edge-DNS resolves it from inside Google's network and returns an answer.

So I agree this isn't a hot path on a healthy client — that's why the PR is framed as "low-risk perf" rather than a fix. It makes an existing cold-but-not-empty fallback cheaper, doesn't promote DNS to a first-class Apps Script concern, and `ENABLE_EDGE_DNS_CACHE = false` reverts to the pre-PR behavior verbatim.

Issue #719, opened 2026-05-04
Need EXCLUDE_SNI_REWRITE_SUBDOMAINS_LIST for app_script mode


TKZ700:
Hi, thank you for the recent changes. The setup works perfectly in full_node mode with an exit node.

However as i mentioned in #701 , when using the Google Script -> CF Worker flow (i.e., running the app in app_script mode), Google hosts cannot be proxied as expected.

The root cause:
In app_script mode, all Google subdomains have their SNI rewritten by default. This becomes problematic when trying to reach Google AI Studio because we are connecting to it with our real IP from Iran which will be blocked. so we can bypass the sanctions by connecting to it through app_script proxy, not directly from the Google Script itself, but when a Cloudflare Worker is used as the endpoint of the chain (via Code.cfw.gs).

To successfully connect to Google AI Studio through the app_script -> CF Worker chain, we need to bypass SNI rewriting for certain Google subdomains, so they can be properly proxied and thus circumvent sanctions.

Proposed fix:
Add an EXCLUDE_SNI_REWRITE_SUBDOMAINS_LIST configuration option for app_script mode. This would allow users to specify addresses like Google AI Studio, ensuring they are proxied through the Google Script -> CF Worker pipeline without SNI modification. A similar strategy to what is being used for youtube_through_relay option, but as a list for EXCLUDE SNI REWRITE SUBDOMAINS.

Looking forward to your thoughts.

o0Fara0o:
سلام من نمیتونم سرور رایگان یا val. رو بسازم ...امکانش هست شما برام انجام بدین ؟ مثلا یوزر و پسورد رو بهتون بدم زحمتشو بکشید یا هر روش دیگری که ممکنه  ....

o0Fara0o:
سلام من نمیتونم سرور رایگان یا val. رو بسازم ...امکانش هست شما برام انجام بدین ؟ مثلا یوزر و پسورد رو بهتون بدم زحمتشو بکشید یا هر روش دیگری که ممکنه  ....

therealaleph:
Reviewed via Anthropic Claude.

@TKZ700 — feature request makes sense. Concept: `disable_sni_rewrite_hosts: Vec<String>` config field that takes precedence over the built-in `SNI_REWRITE_SUFFIXES` list, forcing matched hosts through the Apps Script relay (→ CF Worker chain) instead of the GFE SNI-rewrite tunnel.

[...]

Issue #464, opened 2026-04-29
The Youtube 59 second problem possible workaround


alita8606:
When using "Send Youtube through relay" the relay makes youtube in to thinking the device is mobile

> device: 
>     brand: "danger"
>     browser: "Access Netfront"
>     browserVersion: "3.0"
>     interfaceName: "MWEB"
>     interfacePlayerType: "UNIPLAYER"
>     interfaceVersion: "2.20260427.09.00"
>     model: "hiptop"
>     os: "Hiptop OS"
>     platform: "MOBILE"

with a lot of reloading sometimes the part that detects the device, doesn't work properly and shows the desktop version in that case the video plays smoothly without that 59 second problem

If there is a way to send user-agent in the relay code i guess it would be fixed

therealaleph:
@alita8606 — this is a really interesting observation and a high-signal report. Walking through what you found:

**Your observation in concrete terms:**

When YouTube traffic goes through Apps Script's `UrlFetchApp.fetch()`, YouTube sees the User-Agent that Apps Script sends (rather than the user's actual browser UA). Apps Script's UA is recognized by YouTube as a low-end mobile profile:

```
device: { brand: "danger", os: "Hiptop OS", model: "hiptop", platform: "MOBILE" }
browser: { browserVersion: "3.0", interfaceName: "MWEB" }
```

This `Hiptop` profile is extreme — Hiptop was a 2002-era T-Mobile feature phone. YouTube's mobile-web (MWEB) path optimizes for very-low-bandwidth devices, which includes the SABR streaming protocol that exhibits the 59-second cliff because Apps Script's response cap matches that path's chunk window.

When YouTube occasionally sends the desktop variant (which you observed via reload), the cliff disappears because the desktop path uses different streaming chunk sizes that don't align with the Apps Script cap.

**Why YouTube treats Apps Script as Hiptop:**

Apps Script's `UrlFetchApp.fetch()` defaults its User-Agent to something like:

```
Mozilla/5.0 (compatible; Google-Apps-Script; beanstalk; +http://script.google.com; id: <hash>) Hiptop/...
```

YouTube's UA-parsing pipeline doesn't recognize this pattern as "modern desktop", falls back to a mobile/feature-phone heuristic, and routes through MWEB.

**The actionable question:** can mhrv-rs / Code.gs override the User-Agent on the upstream call?

[...]

PR #977, opened 2026-05-09
feat(youtube): relay_url_patterns + SABR strip + exit-node-full SNI

dazzling-no-more:
## Summary

Ports three upstream YouTube fixes from the Python project (`masterking32/MasterHttpRelayVPN`) and rewires routing in `Mode::AppsScript` so they cooperate with the existing exit-node, fronting-group, and `youtube_via_relay` features.

- **SABR quality-track strip** (upstream `9b6d03e` + `33db28a`) — strips top-level field-3 entries from `/videoplayback` POST bodies on `*.googlevideo.com` / `*.youtube.com`. Fixes "Response too large" 502s on multi-track segment fetches that exceed Apps Script's `UrlFetchApp` 10 MB cap. Heuristic only fires on segment-fetch shape (field-2 present) so session-init bodies stay intact.
- **`relay_url_patterns` path-pinned relay routing** (upstream `b3b9220`) — new config field. Default `youtube.com/youtubei/` is prepended at startup so YouTube's in-page RPC (where SafeSearch / live-stream gating decisions land) goes through the Apps Script relay, while non-`/youtubei/` paths fall through to a fresh SNI-rewrite TLS connection. Recovers the SafeSearch fix that previously required the heavyweight `youtube_via_relay = true` knob, at ~1% of the quota cost.
- **Exit-node-full SNI override** (upstream `88b2767`) — when `exit_node.mode = "full"` is active in `apps_script` mode, YouTube hosts are pulled out of the SNI-rewrite suffix list so every YT request lands in `DomainFronter::relay` and routes through the second-hop exit node, restoring the documented "every URL routes through the exit node" contract.

## What's in the diff

- New config field `relay_url_patterns: Vec<String>` (with full `Config::validate()` coverage of the `host/path-prefix` shape — empty hosts, RFC 1123 label rules, oversized labels, etc.).
- New `ResolvedRouting` struct that resolves patterns + force-MITM hosts at startup, gated to `Mode::AppsScript` only (Mode::Direct and Mode::Full intentionally inert — neither has a relay path the filter could route through).
- New `forward_via_sni_rewrite_http` helper for non-matching paths on path-pinned hosts. Dials `google_ip:443` with `SNI=front_domain` and sends the real `Host` header; gated to safe methods (GET / HEAD / OPTIONS) to avoid replay risk on POST/PUT/PATCH; gated off entirely when `exit_node.mode = "full"` so the bypass path can't undermine the exit node.
- New `SABR strip + url_host_is_youtube_video_endpoint` host gate so unrelated services exposing `/videoplayback` don't get their bodies rewritten.
- Startup `tracing::warn!`s for: patterns whose host isn't SNI-rewrite-capable (forwarder would return wrong-origin from the Google edge — pattern preserved but force-MITM skipped); patterns dropped because a YT-host pattern conflicts with `youtube_via_relay = true`; fronting-group domains overlapping force-MITM hosts (group dispatch wins, path filter inert).
- Desktop UI (`ConfigStore` round-trip) + Android `MhrvConfig` parity — new `relay_url_patterns` field is round-tripped through both wire formats and the `mhrv-rs://` share/import path. No UI editor (power-user knob like `passthrough_hosts`).
- Forwarder response cap dropped from 200 MB to 32 MB — generous + defensive for the realistic max on this code path (HTML / JS / static assets) while shrinking memory blast radius ~6× on memory-constrained devices (OpenWRT / Android). Streaming TODO preserved as a proper followup.

## Test plan

- [x] `cargo build --bins` clean
- [x] `cargo test --lib` — 268 passed (67 new); covers SABR strip parsing, host gate, URL pattern + force-MITM matchers (incl. trailing-dot, port-in-authority, case), SNI-capable filter, RFC 1123 host validation, `ResolvedRouting` per-mode behavior (AppsScript / Direct / Full), exit-node-full + user-pattern interactions, fronting-group precedence, forwarder request rebuilding (chunked → fresh `Content-Length`, hop-by-hop drop, port handling, POST empty-body framing).

### DIFF
--- a/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
@@ -160,11 +160,12 @@ data class MhrvConfig(
     val youtubeViaRelay: Boolean = false,
 
     /**
-     * SABR quality-track strip kill-switch (Rust `sabr_strip`).
-     * Default true. See `src/config.rs` `sabr_strip` for the trade-off
-     * and when to flip — Android-side is just round-trip plumbing.
+     * SABR quality-track strip — opt-in (Rust `sabr_strip`, default
+     * false after #977 testing). See `src/config.rs` `sabr_strip` for
+     * the full reasoning and when to flip on. Android-side is just
+     * round-trip plumbing.
      */
-    val sabrStrip: Boolean = true,
+    val sabrStrip: Boolean = false,
 
     /**
      * Path-pinned relay routing (Rust `relay_url_patterns`).
@@ -256,10 +257,10 @@ data class MhrvConfig(
             put("tunnel_doh", tunnelDoh)
             put("block_doh", blockDoh)
             if (youtubeViaRelay) put("youtube_via_relay", true)
-            // sabr_strip default is true on the Rust side; emit only
-            // when the user has explicitly disabled it so unchanged
-            // configs stay clean. #977 kill-switch.
-            if (!sabrStrip) put("sabr_strip", false)
+            // sabr_strip default is false on the Rust side (opt-in
+            // after #977); emit only when the user has explicitly
+            // enabled it so unchanged configs stay clean.
+            if (sabrStrip) put("sabr_strip", true)
             // Trim/drop-empty/dedupe before serializing — same pattern
             // as bypass_doh_hosts. Skip the key entirely when the user
             // hasn't added any extras so we don't leak an empty array
@@ -495,7 +496,7 @@ object ConfigStore {
             tunnelDoh = obj.optBoolean("tunnel_doh", true),
             blockDoh = obj.optBoolean("block_doh", true),
             youtubeViaRelay = obj.optBoolean("youtube_via_relay", false),
-            sabrStrip = obj.optBoolean("sabr_strip", true),
+            sabrStrip = obj.optBoolean("sabr_strip", false),
             bypassDohHosts = obj.optJSONArray("bypass_doh_hosts")?.let { arr ->
                 buildList { for (i in 0 until arr.length()) add(arr.optString(i)) }
             }?.filter { it.isNotBlank() }.orEmpty(),
--- a/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
+++ b/android/app/src/main/java/com/therealaleph/mhrv/ConfigStore.kt
@@ -160,27 +160,18 @@ data class MhrvConfig(
     val youtubeViaRelay: Boolean = false,
 
     /**
-     * Path-pinned relay routing (Rust `relay_url_patterns`, upstream
-     * commit b3b9220). Each entry is a `host/path-prefix` (no scheme,
-     * lowercase) — paths matching go through the Apps Script relay,
-     * non-matching paths on the same host fall through to a direct
-     * SNI-rewrite HTTP forward (saving Apps Script quota).
-     *
-     * The Rust side prepends the default `youtube.com/youtubei/`
-     * pattern at startup, with two suppression gates:
-     *   - `youtubeViaRelay = true` (full YT through relay → filter is
-     *     redundant)
-     *   - exit-node "full" mode (every URL must route through the
-     *     second-hop exit node → filter would bypass it)
-     * Plus, when either gate is active, user-supplied patterns whose
-     * host overlaps `YOUTUBE_RELAY_HOSTS` (youtube.com, youtu.be,
-     * youtube-nocookie.com, youtubei.googleapis.com) are dropped at
-     * startup with a warning, since they would partially defeat the
-     * full-relay contract.
-     *
-     * This Android-side field is for *additional* user entries only —
-     * round-tripped through config.json so a hand-edited extension
-     * survives a save. No UI editor (power-user knob).
+     * SABR quality-track strip kill-switch (Rust `sabr_strip`).
+     * Default true. See `src/config.rs` `sabr_strip` for the trade-off
+     * and when to flip — Android-side is just round-trip plumbing.
+     */
+    val sabrStrip: Boolean = true,
+
+    /**
+     * Path-pinned relay routing (Rust `relay_url_patterns`).
+     * See `src/config.rs` `relay_url_patterns` for the full semantics —
+     * suppression gates, default pattern, host-overlap rules. This
+     * Android-side field is for *additional* user entries only,
+     * round-tripped so a hand-edited list survives Save.
      */
     val relayUrlPatterns: List<String> = emptyList(),
 
@@ -265,6 +256,10 @@ data class MhrvConfig(
             put("tunnel_doh", tunnelDoh)
             put("block_doh", blockDoh)
             if (youtubeViaRelay) put("youtube_via_relay", true)
+            // sabr_strip default is true on the Rust side; emit only
+            // when the user has explicitly disabled it so unchanged
+            // configs stay clean. #977 kill-switch.
+            if (!sabrStrip) put("sabr_strip", false)
             // Trim/drop-empty/dedupe before serializing — same pattern
             // as bypass_doh_hosts. Skip the key entirely when the user
             // hasn't added any extras so we don't leak an empty array
@@ -385,6 +380,7 @@ object ConfigStore {
         if (cfg.tunnelDoh != defaults.tunnelDoh) obj.put("tunnel_doh", cfg.tunnelDoh)
         if (cfg.blockDoh != defaults.blockDoh) obj.put("block_doh", cfg.blockDoh)
         if (cfg.youtubeViaRelay != defaults.youtubeViaRelay) obj.put("youtube_via_relay", cfg.youtubeViaRelay)
+        if (cfg.sabrStrip != defaults.sabrStrip) obj.put("sabr_strip", cfg.sabrStrip)
         val cleanBypassDohHosts = cfg.bypassDohHosts
             .map { it.trim() }
             .filter { it.isNotEmpty() }
@@ -499,6 +495,7 @@ object ConfigStore {
             tunnelDoh = obj.optBoolean("tunnel_doh", true),
             blockDoh = obj.optBoolean("block_doh", true),
             youtubeViaRelay = obj.optBoolean("youtube_via_relay", false),
+            sabrStrip = obj.optBoolean("sabr_strip", true),
             bypassDohHosts = obj.optJSONArray("bypass_doh_hosts")?.let { arr ->
                 buildList { for (i in 0 until arr.length()) add(arr.optString(i)) }
             }?.filter { it.isNotBlank() }.orEmpty(),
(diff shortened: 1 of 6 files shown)


therealaleph:
Reviewed via Anthropic Claude. Read the PR body + structural diff.

Big PR, three independent strands bundled. Let me address each:

**1. `relay_url_patterns` — generic URL-pattern routing knob.**
Long-standing ask (related to #719's `EXCLUDE_SNI_REWRITE_SUBDOMAINS_LIST`). A user-configurable allowlist that overrides the default suffix-routing logic is the right shape. The pattern syntax (regex vs glob vs prefix) is what determines whether this is friendly or footgun-y — leaning toward prefix-with-wildcard for the surface area, but I'll re-read the implementation choice.

**2. SABR query-string strip.**
Direct attack on the YouTube 59s problem (#464) — strips `&sabr=1&rqh=1` from `googlevideo.com` URLs before they hit Apps Script. If the empirical claim ("YouTube falls back to a non-SABR streaming path that doesn't have the 59s cliff") holds across users, this is the breakthrough we've been after. Risk: if Google starts rejecting requests with stripped sabr params (server-side validation), it could break video playback entirely for all users at once. Wants a config flag to opt out — looks like the implementation exposes one. Good.

**3. exit-node SNI for full mode.**
Lets full-mode users send `Origin: https://chatgpt.com` to their exit-node so the chained downstream destination's SNI matches what the browser thinks it's doing. Solves a class of "exit-node returns 421 Misdirected Request" bugs.

**+2722 / -31** across UI + config + domain_fronter + proxy_server + Android Kotlin — substantial. The structural changes look well-scoped (each strand is in its own module), but I want to sanity-check the regression surface.

**Plan**: leaving open for **5–7 days community testing** (same path as #903 / #359). Specifically asking testers to verify:
1. **YouTube videos play past 59 seconds** — the headline win. Try 3+ videos of varying length (>5 min, >15 min, full-length livestream).
2. **Subtitles still load** (r0ar in #962 reported subtitles broke under `force_http1=true`; want to confirm the SABR strip path doesn't have a similar side-effect).
3. **No regression on non-YouTube apps_script browsing** — login flows on Google, Reddit, X, etc.
4. **Exit-node-full SNI** — try chatgpt.com / claude.ai through your exit-node and confirm responses are normal HTML, not SNI-mismatch errors.

Code-wise — will do a closer read on:
- The pattern matcher's complexity guarantees (don't want a regex DOS via user-supplied pattern).
- Whether `relay_url_patterns` shadows the existing `passthrough_hosts` semantics or composes with them.
- Android Kotlin `ConfigStore.kt` round-trip for the new config fields (this is where past PRs have regressed — see v1.9.11/v1.9.12 build failures).

Verified locally that build is clean + tests pass on top of v1.9.18:
- `cargo build --bins --lib`: clean
- `cargo test --lib --release`: 208/208

Thanks @dazzling-no-more — this is exactly the right strand of work to be cutting through. Will pre-test myself on the SABR strip claim and report back here.

w0l4i (approved):
Great commit, clever move !
keep it going champ 💪

unacoder:
> YouTube videos play past 59 seconds — the headline win. Try 3+ videos of varying length (>5 min, >15 min, full-length livestream).

On youtube only few seconds are buffered, at 1.7x or 2x playback it constantly hitting stops and makes playback unusable (testing at 360p/480p, `"youtube_via_relay": false,`), this is fine in main.

One thing I saw in my testing, switching between main and this branch causes youtube buffer to drop completely, maybe this change makes buffering useless somehow? Seems like only one videoplayback request is buffered

Testing with `FullMode` and `"youtube_via_relay": true,`, buffering works as expected

> Subtitles still load (r0ar in https://github.com/therealaleph/MasterHttpRelayVPN-RUST/issues/962 reported subtitles broke under force_http1=true; want to confirm the SABR strip path doesn't have a similar side-effect).

Subtitles are fine

> No regression on non-YouTube apps_script browsing — login flows on Google, Reddit, X, etc.

No changes in everyday usage, about login flows I was unable to login to stackexchange sites or github neither in this branch nor on main. (Tested only `AppScript`), I think this was the case from the early versions.

> Exit-node-full SNI — try chatgpt.com / claude.ai through your exit-node and confirm responses are normal HTML, not SNI-mismatch errors.

Not tested

dazzling-no-more:
@unacoder huge thanks for the detailed testing — the buffer-drains-faster-than-it-fills behavior at 1.7×–2× playback is exactly the kind of thing that wouldn't have shown up in a quick smoke test.

Pushed a follow-up commit with a kill-switch:

```json
{ "sabr_strip": false }
```

Default stays `true` (matches upstream behavior + keeps the "no more 502s on multi-track segments" win for users who hit *that* bug). Flipping to `false` reverts to pre-strip behavior — Google bundles multi-quality responses, occasional 502s on long segments, player handles by dropping quality.

To help isolate whether the SABR strip is actually the cause, would you mind retesting in `apps_script` mode with **two** more configs?

1. `youtube_via_relay: false` + `sabr_strip: false` — keeps the new path-routing on, disables the SABR strip. Buffering normal here ⇒ SABR strip is the culprit.
2. `youtube_via_relay: true` + `sabr_strip: true` — keeps SABR strip on, disables the path-routing (full YT through relay). Buffering normal here ⇒ the new SNI-rewrite forwarder is the culprit (no connection pool / fresh TLS handshake per asset request).

If both are normal but the default config still breaks, it's the interaction. Either result narrows down the next followup — we have a "keep-alive on the forwarder" change queued and an "opt-in instead of opt-out for the default pattern" change queued, but we'd rather pick the right one.

Subtitles + non-YouTube unchanged is good news — confirms the SABR strip's host gate is correctly tight (only fires on `*.googlevideo.com` / `*.youtube.com`) and the routing changes don't bleed into other sites.

unacoder:
> 1. `youtube_via_relay: false` + `sabr_strip: false `— keeps the new path-routing on, disables the SABR strip. Buffering normal here ⇒ SABR strip is the culprit.

Buffering works as expected

> 2. `youtube_via_relay: true` +` sabr_strip: true` — keeps SABR strip on, disables the path-routing (full YT through relay). Buffering normal here ⇒ the new SNI-rewrite forwarder is the culprit (no connection pool / fresh TLS handshake per asset request).

Short buffer (as described in previous comment), this buffer issue is not time related, even after a full minute youtube buffer doesn't load more even though logs show googlevideo is requested again.

I tested both with default behavior `Auto (360p)`, and setting quality to exact 360p, same result

Also tested in chrome too, same behavior (all other was with firefox)

One important note here, since `youtube_via_relay` doesn't load youtube page fully initial page load is done without that config, it's a 1 hour video and I moved the seeker at random points each test. Loading `m.youtube.com` with second config loads interface still no buffering, can't test `m.youtube.com` with first config (redirects to main site anyway).

Another observation when I left it to buffer with both config set to true, it requested semi exact url at `googlevideo.com` 11 times with only difference between those calls being the number for `rn` query param incrementing: `&rn=34&`, the fact that it requests 11 times without buffering seems odd (although might be artifact of the loading interface one way and loading video other way). With both set to false it requests 3 times, with each buffer extends a bit.

Bellow is stats for nerds using chrome, looking at Buffer health, those green/yellow section are when both false is set, orange/yellow and also black section are when using second config (both true)

www youtube com_watch_app=desktop v=SnUYaIUpQec

P.s couldn't take more screenshots (linux updates messed up things)

dazzling-no-more:
@unacoder thanks — your two-test split was decisive. Both directions:

- **Test 1** (forwarder ON, SABR OFF) → buffering normal. **Forwarder is innocent.** No keep-alive/H2 work needed on it.
- **Test 2** (forwarder OFF, SABR ON) → short buffer, 11 retries with `rn=` incrementing. **SABR strip is the regression source.**

The 11-retries-with-rn-incrementing signature was the giveaway. Our hypothesis: the player at 360p sends a single field-3 quality-track selector. The original "strip all field-3" rule turned that into a request with **zero tracks selected** — googlevideo answered with an empty body, the player retried indefinitely, never advancing the buffer.

Pushed a refined heuristic that diverges from upstream Python's strip-all rule:

> **Keep the first field-3 entry, strip the rest. Only fire when there are 2+ field-3 entries AND a field-2 byte-range is present.**

What this means in practice:

- **Single-track requests** (one field-3 — the common case at low/medium quality) pass through unchanged. googlevideo always has a track selected → no empty responses → buffer fills normally.
- **Multi-track requests** (2+ field-3 — high-quality bundling that hits the 10 MB cap) still get capped to one track, so the original "Response too large" 502 fix is preserved.
- **Session-init requests** (no field-2) untouched, same as before.

Default stays `sabr_strip: true` because the new heuristic should be safe for both cases. The kill-switch (`"sabr_strip": false`) remains as an escape hatch if you find a body shape we didn't anticipate.

Could you retest in `apps_script` mode with **default config** (no overrides — both `youtube_via_relay` and `sabr_strip` at their defaults)? That's the path that was broken on `main` of this branch, and the path most users will land on. If buffering is normal, we ship; if the same `rn=` retry signature comes back, we have more digging to do (likely captures of the actual `/videoplayback` request body) and we can fall back to flipping `sabr_strip` default to `false` for safety.

The screenshot and the `googlevideo.com` request count breakdown (3 vs 11) were extremely useful — that quantitative diff is what made the failure mode legible.

unacoder:
Same results, I don't think quality is the culprit here, even at 1080p60 (1x speed) first buffers just fine, stats get green (albeit slowly) but with second config it never  gets past 20s buffer. 

Tested with example.config only `script_id` and `auth_key` changed (buffer issue persists), also ran those two configs before (same results as before).

IMHO `sabr_strip` should be false by default. To me it seems like new request replaces entire buffer and is not appending, at the same time number of requests to googlevideo is higher.

Also checked other repo's telegram channel and found nothing related to sabr

dazzling-no-more:
@unacoder you were right, flipped it. Pushed the change:

- `sabr_strip` now defaults to **`false`**
- `default_sabr_strip()` returns `false`; both desktop UI and Android `ConfigStore` defaults match; JSON serialization only emits the key when explicitly enabled
- The keep-first heuristic stays as the opt-in behaviour — strictly less aggressive than upstream's strip-all, so users who specifically hit "Response too large" 502s on long-form 1080p+ playback get the safer flavour. Most users should leave it off.

What I think went wrong, putting the data together:

The two failed iterations (strip-all, keep-first) plus your "new request replaces entire buffer" observation point at field-3 entries not being homogeneous quality-track selectors the way upstream Python's comment claims. They likely encode multiple request facets — audio track, video track, init-segment refs, per-stream byte-range modifiers — that the player needs as a set, not as redundant variants. Stripping any subset produces responses the player can't splice into its existing buffer, so it discards what it had and starts over (matches your "replaces entire buffer" exactly).

That also explains why upstream's Telegram channel is silent on this. Iranian users on the upstream may be on a different YouTube backend (region-routed CDN) or a different player code path where field-3 is genuinely redundant. We can't replicate their conditions, and the safe move is to stop guessing.

If/when someone can capture a real `/videoplayback` POST body (paste-bin friendly) with proto reflection, we can decode field-3 and design a heuristic that's actually correct. Until then, default-off ships the unbroken-playback common case.

Re: the "buffer replacement, not append" symptom — that was exactly the kind of qualitative observation that pinned down the failure mode. The keep-first iteration was a guess; you confirmed the guess was wrong before we'd shipped it to anyone else. Thanks.

PR #1040, opened 2026-05-10
perf: skip H2 for full-tunnel batch requests

yyoyoian-pixel:
## Summary

Full-tunnel batches already coalesce N ops into one HTTP request — H2 stream multiplexing has nothing to multiplex. This PR skips the H2 path for `tunnel_batch_request_to` and keeps the H1 pool at full capacity.

H2 multiplexing remains active for **relay mode** (non-full) where each browser request is a separate HTTP call that genuinely benefits from stream multiplexing.

## Problem

H2 for tunnel batches introduced three regressions vs v1.9.14:

1. **Long-poll stalls**: idle polls complete at 16–17s (LONGPOLL_DEADLINE + network latency) instead of timing out at 10s on H1. Each poll holds an Apps Script execution slot 60% longer.
2. **Silent batch drops**: `RequestSent::Maybe` failures drop the entire batch with no retry — a failure mode H1 doesn't have.
3. **Pool starvation**: `POOL_MIN_H2_FALLBACK` trims the H1 pool from 8→2 once H2 connects, but tunnel batches still use H1 and need the full pool.

## A/B test results (Pixel 6 Pro, 30 batch samples each)

| Metric | H2 (stock v1.9.20) | H1 (this PR) | v1.9.14 (baseline) |
|---|---|---|---|
| 16–17s batches | **8–10/30** | **0/30** | **0/30** |
| 10s timeouts | 0 | 4/30 | 5/30 |
| Active RTTs | 1.4–2.4s | 1.3–2.2s | 1.4–2.3s |

This PR restores v1.9.14 tunnel performance while keeping all v1.9.15+ improvements (H2 for relay, zero-copy mux, block DoH/QUIC, TLS pool tuning).

## Changes

- `tunnel_batch_request_to`: remove H2 try/fallback/NonRetryable block, go straight to H1 pool `acquire()` (-54 lines)
- `run_pool_refill`: always maintain `POOL_MIN` (8). Remove `POOL_MIN_H2_FALLBACK` (2) trim that starved tunnel batches (-12 lines)

## Test plan

- [x] `cargo check` clean
- [x] A/B tested on Pixel 6 Pro — v1.9.14 performance restored
- [x] Four separate sample runs confirm zero 16-17s stalls with H1
- [x] H2 still active for relay mode (not affected)

### DIFF
--- a/src/domain_fronter.rs
+++ b/src/domain_fronter.rs
@@ -983,16 +983,13 @@ impl DomainFronter {
         }
     }
 
-    /// Background loop that keeps the h1 fallback pool warm.
+    /// Background loop that keeps the h1 pool warm.
     ///
-    /// Target depends on whether the h2 fast path is active:
-    /// - h2 disabled (or peer refused ALPN h2): keep `POOL_MIN` (8)
-    ///   sockets so the per-request acquire never pays a cold handshake
-    ///   — the pre-h2 default behavior.
-    /// - h2 active: keep just `POOL_MIN_H2_FALLBACK` (2). All real
-    ///   traffic rides the multiplexed h2 connection; the h1 pool only
-    ///   exists to cover the case where h2 dies and we need to fall
-    ///   back instantly without a cold handshake.
+    /// Always maintains `POOL_MIN` (8) connections. Full-tunnel mode
+    /// uses the h1 pool for all batch traffic (h2 is skipped for
+    /// tunnel batches), so the pool must stay at full capacity
+    /// regardless of h2 status. Relay mode also benefits from a warm
+    /// pool as h1 fallback.
     ///
     /// A connection only counts toward the minimum if it has at least
     /// 20 s of TTL remaining — nearly-expired entries don't help.
@@ -1000,7 +997,6 @@ impl DomainFronter {
     /// and opens replacements one at a time so there's no burst.
     pub async fn run_pool_refill(self: Arc<Self>) {
         const MIN_REMAINING_SECS: u64 = 20;
-        const POOL_MIN_H2_FALLBACK: usize = 2;
         loop {
             tokio::time::sleep(Duration::from_secs(POOL_REFILL_INTERVAL_SECS)).await;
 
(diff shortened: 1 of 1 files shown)


therealaleph:
Reviewed via Anthropic Claude. Verified locally:
- `cargo test --lib --release`: 209/209 ✅
- `cargo build --release --features ui --bin mhrv-rs-ui`: clean ✅

The architectural reasoning is sound: tunnel batches are already-coalesced single HTTP requests where H2 stream multiplexing has no value (nothing to multiplex). H2 multiplexing genuinely shines on relay mode (one HTTP call per browser request, dozens parallel) — and you've explicitly kept H2 there. The A/B data on Pixel 6 Pro is concrete (8-10/30 stalls → 0/30).

Two things I want to verify before merging:

**1. Interaction with PR #1029 (just merged in v1.9.20)**: that PR was @rezaisrad's fix for the 3-week #924 cluster — same `warm()` race in domain_fronter.rs that this PR also touches. Their fix parallelized h1 prewarm with h2 handshake and added `H2Cell.dead` AtomicBool for synchronous dead-cell detection. With this PR removing the h2 path for tunnel_batch_request_to entirely, the `H2Cell.dead` flag stops getting consulted on the tunnel batch path — that's fine (no regression) but the dead-cell logic now only covers relay mode. Worth a one-liner comment in the changelog noting this.

**2. Cross-check vs #962 (r0ar's data)**: that issue's controlled A/B showed h2 enabled was strictly better than `force_http1: true` for **apps_script mode**, not Full mode. Your PR scopes the skip to `tunnel_batch_request_to` only (Full mode batch path), so r0ar's apps_script users keep h2 — same trade-off as your "H2 still active for relay mode" claim. Just confirming this matches the diff.

Plan: leaving open for **3-5 days** community testing. Specifically asking testers:
1. Cold-start Full mode after `mhrv-rs` restart — does v1.9.20+this PR feel snappier than stock v1.9.20?
2. Long sustained browse session (>30 min) — any new stalls or disconnect patterns?
3. Mixed apps_script + Full mode (run both deployments via separate `script_id`s): does relay mode latency regress?

If 2-3 testers confirm "feels like v1.9.14 again" without surprise regressions, I'll squash-merge as v1.9.21.

Excellent perf work — the empirical bisect-into-h2-overhead is the right framing for these kinds of low-level transport regressions.

---

PR #1085, opened 2026-05-12
fix(relay): stream range-parallel downloads larger than Apps Script's 50 MiB cap

### LINKED ISSUE #1042: H2 resurrection
Why not H2 until Gas then goes http 1.1 to origin ?

H2 performance maintain + H1 reliability For Full mode

Client H2 > Gas - (H2 to Http1 transform) > Http 1.1 With encryption over ssl to vps > Nginx / caddy (ssl termination + reverse proxy) to node tunnel

dazzling-no-more:
## Summary
- Fixes #1042: range-capable downloads larger than ~50 MiB through the Apps Script relay returned `504 Relay timeout — Apps Script unresponsive` instead of the file. The 104 MiB v2rayN DMG in the reported logs is the canonical repro.
- Root cause: `relay_parallel_range` capped the stitched response at 64 MiB and fell back to a single `relay()` for anything larger. Single-GET routes through Apps Script's ~50 MiB response ceiling, so Apps Script killed the script mid-execution and we hung for the full 25 s relay timeout before returning 504.
- Fix: convert `relay_parallel_range` into a writer-based API that streams large files chunk-by-chunk to the client socket. Each chunk is still one ≤256 KiB Apps Script call (well under the 50 MiB cap); only the host-side buffering changes. A backward-compatible `Vec<u8>` wrapper preserves the pre-1.9.23 API surface.

## What changed
- **`src/domain_fronter.rs`**
  - New writer-based API `relay_parallel_range_to(&mut writer, …, transform_head)`. Streams the synthetic `200 OK` head + probe body + remaining chunks straight to a `tokio::io::AsyncWrite` target.
  - Old `relay_parallel_range -> Vec<u8>` retained as a thin compatibility wrapper for external library consumers.
  - Three-way dispatch via `RangeDispatch { Buffered, Stream, FallbackSingleGet, RejectTooLarge }` and the pure `dispatch_range_response(total, streaming_allowed)` predicate:
    - **`Buffered`** — `total ≤ APPS_SCRIPT_BODY_MAX_BYTES` (40 MiB) on either surface, plus the 40–64 MiB band for the wrapper. Existing stitch + single-GET fallback path; fully recovers on chunk failure.
    - **`Stream`** — writer API above 40 MiB. Streams; chunk failure flushes the committed prefix and returns `Err` so the wire body's `Content-Length` mismatch tells download clients to resume via `Range`.
    - **`FallbackSingleGet`** — wrapper above 64 MiB. Falls back to `self.relay()`, matching the pre-1.9.23 cliff.
    - **`RejectTooLarge`** — writer API above 16 GiB. Refuses with a 502; bounds worst-case Apps Script quota drain from a hostile origin advertising an absurd `Content-Range` total.
  - Lazy `plan_remaining_ranges` (via `std::iter::from_fn` + `saturating_*`): range planning is `O(1)` memory regardless of advertised total, so even a `u64::MAX` total no longer drives a ~6 GB `Vec<(u64, u64)>` allocation.
  - Free helpers extracted for testability: `assemble_200_head`, `write_response_with_head_transform`, `stream_chunks_to_writer`, `stream_range_response_to`, `VecAsyncWriter`.
  - Streaming flushes after head + probe (first-byte latency) and before returning a chunk-validation `Err` (TLS buffer drains so clients see the partial body needed for Range-resume).
- **`src/proxy_server.rs`**
  - MITM HTTPS and plain-HTTP call sites updated to use `relay_parallel_range_to` with a CORS-aware `transform_head` closure.
  - Extracted `inject_cors_into_head` (head-only variant of `inject_cors_response_headers`, marked `pub(crate)`) so the streaming path can rewrite ACL headers before the body has been assembled.

## Test plan
- [x] `cargo test --lib` — 227 passing (15 new tests for dispatch enum, lazy range planning, head assembly, head transform, streaming writer, flush behavior, and a cross-module CORS-into-streaming integration test).
- [x] `cargo clippy --lib` — clean for touched code; only pre-existing warnings elsewhere.
- [ ] Manual download repro from #1042: fetch the 104 MiB v2rayN macOS DMG (`https://github.com/2dust/v2rayN/releases/download/7.19.5/v2rayN-macos-64.dmg`) through the proxy and confirm the file completes with matching SHA-256 instead of a 504.

### DIFF
--- a/src/domain_fronter.rs
+++ b/src/domain_fronter.rs
@@ -178,6 +178,16 @@ const APPS_SCRIPT_BODY_MAX_BYTES: u64 = 40 * 1024 * 1024;
 /// through.
 const MAX_STREAMED_RANGE_BYTES: u64 = 16 * 1024 * 1024 * 1024;
 
+/// Byte interval between `range-parallel-stream` progress log lines.
+/// Large downloads through the streaming branch otherwise look stuck
+/// in the logs (one "starting N chunks" line at the top, nothing
+/// until completion or failure). At 16 MiB intervals the operator sees
+/// ~6 lines per 100 MiB and ~64 lines per 1 GiB — useful pace at the
+/// ~1.4 MB/s typical through-relay throughput, and quiet enough that
+/// even a 16 GiB file won't drown the log (~1024 progress lines over
+/// the multi-hour download). Per user feedback on PR #1085.
+const STREAM_PROGRESS_LOG_INTERVAL_BYTES: u64 = 16 * 1024 * 1024;
+
 /// Hard ceiling on the buffered stitch buffer's `Vec::with_capacity(total)`
 /// allocation. Two roles:
 ///
@@ -3582,6 +3592,7 @@ async fn stream_chunks_to_writer<W, S>(
     writer: &mut W,
     head: &[u8],
     probe_body: &[u8],
+    total: u64,
     fetches: S,
     url_for_log: &str,
 ) -> std::io::Result<()>
@@ -3609,9 +3620,44 @@ where
     writer.flush().await?;
     futures_util::pin_mut!(fetches);
 
+    // Progress accounting: bytes emitted as wire body so far (the
+    // probe body, plus every successfully-written chunk). The head
+    // doesn't count — it's protocol framing, not body progress.
+    // `next_progress_log_at` is the next body-byte threshold at
+    // which we emit a progress line, advanced past the current
+    // count each time so a single large chunk crossing multiple
+    // intervals only logs once.
+    let mut body_bytes_emitted: u64 = probe_body.len() as u64;
+    let mut next_progress_log_at: u64 = STREAM_PROGRESS_LOG_INTERVAL_BYTES;
+
     while let Some((s, e, chunk_result)) = fetches.next().await {
         match chunk_result {
-            Ok(c) => writer.write_all(&c).await?,
+            Ok(c) => {
+                writer.write_all(&c).await?;
+                body_bytes_emitted = body_bytes_emitted.saturating_add(c.len() as u64);
+                if body_bytes_emitted >= next_progress_log_at {
+                    // Percentage is well-defined here: streaming
+                    // branch is only reached for total >
+                    // APPS_SCRIPT_BODY_MAX_BYTES (≥ 40 MiB), so the
+                    // divisor is never zero.
+                    let pct = (body_bytes_emitted * 100) / total;
+                    tracing::info!(
+                        "range-parallel-stream: {}/{} MiB ({}%) emitted for {}",
+                        body_bytes_emitted / (1024 * 1024),
+                        total / (1024 * 1024),
+                        pct,
+                        url_for_log,
+                    );
+                    // Advance to the next interval past the current
+                    // count — a chunk much larger than the interval
+                    // (shouldn't happen at 256 KiB chunks, but defend
+                    // against future tuning) skips intermediate
+                    // thresholds rather than firing N log lines back
+                    // to back.
+                    next_progress_log_at = body_bytes_emitted
+                        .saturating_add(STREAM_PROGRESS_LOG_INTERVAL_BYTES);
+                }
+            }
             Err(reason) => {
                 tracing::warn!(
                     "range-parallel-stream: invalid chunk {}-{} for {} ({}); truncating response",
@@ -3664,7 +3710,7 @@ where
 {
     let head = assemble_200_head(probe_resp_headers, total);
     let head = transform_head(&head);
-    stream_chunks_to_writer(writer, &head, probe_body, chunks_stream, url_for_log).await
+    stream_chunks_to_writer(writer, &head, probe_body, total, chunks_stream, url_for_log).await
 }
 
 /// Tiny adapter that lets `relay_parallel_range_to` write into a
@@ -5498,6 +5544,7 @@ hello";
             &mut VecAsyncWriter(&mut buf),
             head,
             probe,
+            10,
             fetches.map(|x| x),
             "https://example.test/file",
         )
@@ -5726,6 +5773,7 @@ hello";
             &mut writer,
             head,
             probe,
+            12,
             fetches.map(|x| x),
             "https://example.test/file",
         )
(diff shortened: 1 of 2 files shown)


w0l4i:
> ## Summary
> * Fixes [H2 resurrection #1042](https://github.com/therealaleph/MasterHttpRelayVPN-RUST/issues/1042): range-capable downloads larger than ~50 MiB through the Apps Script relay returned `504 Relay timeout — Apps Script unresponsive` instead of the file. The 104 MiB v2rayN DMG in the reported logs is the canonical repro.
> * Root cause: `relay_parallel_range` capped the stitched response at 64 MiB and fell back to a single `relay()` for anything larger. Single-GET routes through Apps Script's ~50 MiB response ceiling, so Apps Script killed the script mid-execution and we hung for the full 25 s relay timeout before returning 504.
> * Fix: convert `relay_parallel_range` into a writer-based API that streams large files chunk-by-chunk to the client socket. Each chunk is still one ≤256 KiB Apps Script call (well under the 50 MiB cap); only the host-side buffering changes. A backward-compatible `Vec<u8>` wrapper preserves the pre-1.9.23 API surface.
> 
> ## What changed
> * **`src/domain_fronter.rs`**
>   
>   * New writer-based API `relay_parallel_range_to(&mut writer, …, transform_head)`. Streams the synthetic `200 OK` head + probe body + remaining chunks straight to a `tokio::io::AsyncWrite` target.
>   * Old `relay_parallel_range -> Vec<u8>` retained as a thin compatibility wrapper for external library consumers.
>   * Three-way dispatch via `RangeDispatch { Buffered, Stream, FallbackSingleGet, RejectTooLarge }` and the pure `dispatch_range_response(total, streaming_allowed)` predicate:
>     
>     * **`Buffered`** — `total ≤ APPS_SCRIPT_BODY_MAX_BYTES` (40 MiB) on either surface, plus the 40–64 MiB band for the wrapper. Existing stitch + single-GET fallback path; fully recovers on chunk failure.
>     * **`Stream`** — writer API above 40 MiB. Streams; chunk failure flushes the committed prefix and returns `Err` so the wire body's `Content-Length` mismatch tells download clients to resume via `Range`.
>     * **`FallbackSingleGet`** — wrapper above 64 MiB. Falls back to `self.relay()`, matching the pre-1.9.23 cliff.
>     * **`RejectTooLarge`** — writer API above 16 GiB. Refuses with a 502; bounds worst-case Apps Script quota drain from a hostile origin advertising an absurd `Content-Range` total.
>   * Lazy `plan_remaining_ranges` (via `std::iter::from_fn` + `saturating_*`): range planning is `O(1)` memory regardless of advertised total, so even a `u64::MAX` total no longer drives a ~6 GB `Vec<(u64, u64)>` allocation.
>   * Free helpers extracted for testability: `assemble_200_head`, `write_response_with_head_transform`, `stream_chunks_to_writer`, `stream_range_response_to`, `VecAsyncWriter`.
>   * Streaming flushes after head + probe (first-byte latency) and before returning a chunk-validation `Err` (TLS buffer drains so clients see the partial body needed for Range-resume).
> * **`src/proxy_server.rs`**
>   
>   * MITM HTTPS and plain-HTTP call sites updated to use `relay_parallel_range_to` with a CORS-aware `transform_head` closure.
>   * Extracted `inject_cors_into_head` (head-only variant of `inject_cors_response_headers`, marked `pub(crate)`) so the streaming path can rewrite ACL headers before the body has been assembled.
> 
> ## Test plan
> * [x]  `cargo test --lib` — 227 passing (15 new tests for dispatch enum, lazy range planning, head assembly, head transform, streaming writer, flush behavior, and a cross-module CORS-into-streaming integration test).
> * [x]  `cargo clippy --lib` — clean for touched code; only pre-existing warnings elsewhere.
> * [ ]  Manual download repro from [H2 resurrection #1042](https://github.com/therealaleph/MasterHttpRelayVPN-RUST/issues/1042): fetch the 104 MiB v2rayN macOS DMG (`https://github.com/2dust/v2rayN/releases/download/7.19.5/v2rayN-macos-64.dmg`) through the proxy and confirm the file completes with matching SHA-256 instead of a 504.

Great commit, clever move !
keep it going champ 💪

w0l4i (approved):

unacoder:
Downloading 169mb file works correctly with this:

```
🗃 Downloads > proxy curl https://www.expressvpn.works/clients/linux/expressvpn-linux-universal-14.1.0.13058_release.run -o express.run
  % Total    % Received % Xferd  Average Speed  Time    Time    Time   Current
                                 Dload  Upload  Total   Spent   Left   Speed
100 169.3M 100 169.3M   0      0  1.15M      0   02:27   02:27          1.31M
🗃 Downloads > ls -lht express.run                                                                      2m27s35ms at 15:12:16
-rw-r--r-- 1 user user 170M May 12 15:12 express.run
🗃 Downloads > gpg --verify expressvpn-linux-universal-14.1.0.13058_release.run.asc express.run                    at 15:13:14
gpg: Signature made Thu 30 Apr 2026 04:33:56 PM +0330
gpg:                using RSA key 1D0B09AD6C93FEE93FDDBD9DAFF2A1415F6A3A38
gpg: Good signature from "ExpressVPN Release <release@expressvpn.com>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 1D0B 09AD 6C93 FEE9 3FDD  BD9D AFF2 A141 5F6A 3A38
```

Speed is capped at 1.4mb/s.

Downloading with firefox works as expected and correctly (gpg verify passed), same speed as curl (or slightly faster).

Couldn't download with aria2, seems this approach won't work with download managers, couldn't test with `yt-dlp` got 429/403 errors. These were extra tests to see how far can it go.

---
Running steam through proxy (this is using different exit nodes and fronting groups for most parts and just trying to test downloads, it won't work with default config):

I was able to download 300mb update, speed was slower than curl, fluctuating a lot, this I think is not a good test for this branch since steam uses a custom logic for downloading (hitting different cache servers and whatnot), included for reference

---
There was no new issues in youtube playback.

P.S there should be some more info logging (any form of progress maybe?)

---
Another test was trying to stop download and see if server properly cancels rest of requests, after stopping a download there was no additional network requests, download/upload speeds return to zero almost immediately. Unsure if app scripts also cancels.

---
A bit of rant unrelated to this PR, wish more people tested PR's here, since full mode is broken/unusable I'm stuck on an older revision, until a fix I can't daily drive newer versions.

dazzling-no-more:
Thanks for the thorough testing — 169 MB with GPG verify is exactly the signal I was hoping for.

Quick notes on each point:

**Speed (~1.4 MB/s):** Expected with the current settings, not a bug. Throughput is bounded by `MAX_PARALLEL × chunk_size / Apps_Script_RTT` ≈ 16 × 256 KiB / 3 s ≈ 1.3 MB/s. Higher parallelism and/or bigger chunks would lift this, but both push closer to Apps Script's per-account quota — happy to tune in a separate PR if there's appetite.

**aria2:** aria2 fires its own parallel `Range` requests. The relay honors caller-sent `Range` headers as-is and skips the parallel-fetch path (so the rest of `aria2`'s logic isn't reentered through it). If aria2 picks a chunk size >~50 MiB you'll hit Apps Script's response cap on the single underlying call. Separate issue from this PR — happy to look at it if you can grab a `--debug` log from aria2 showing the actual Range header it sends.

**yt-dlp 429/403:** Origin-side rate limit, not relay-related.

**Cancellation:** When you cancel client-side, our chunk-fetch futures get dropped, so no more Apps Script calls go out. The Apps Script invocations *already in flight* keep running server-side though — `UrlFetchApp.fetch()` has no cancel API. Usually a handful of chunks finish out and get discarded; doesn't affect quota visibly.

**Progress logging:** Good call, will add a periodic line in a follow-up — something like a chunk-count + bytes-emitted update every N chunks so you can see large downloads making progress in the logs.

Thanks again 🙏

dazzling-no-more:
@unacoder pushed the logging.

Issue #1107, opened 2026-05-12
is full mode good ?


hfjksgk:
i want to buy a vps and deploy full mode . but im not sure if its worth it and i see so many issues saying the speed is slow . can someone who uses it tell me if its good or not and also how do you connect to a vps when iran internet is completly restricted?

NightmareYeager:
no, it dose not work from yesterday

dazzling-no-more:
I just made a PR that fixes this issue.

letmeinbruh13:
> I just made a PR that fixes this issue.

@dazzling-no-more Thank you. Those timeouts were another everyday issue i was about to mention. Telegram specifically suffers the most from them. Sometimes there are too many, so many that telegram stops working even though it might say it's connected. For example, not even a small emoji or a video thumbnail loads, new messages don't come through, and sometimes you can't even send a message, even though the local socks proxy in Telegram says "connected." When it happens, I just leave it as is and try again after a couple of hours.

r-safavi:
@dazzling-no-more can you check issue #620 as well? I can't create a tunnel node using Google Cloud Run and the issue was closed as not planned.

mirbaagheri:
مدل Full Tunnel رو من این چند روز بارها با نسخه های مختلف تست کردم. در نسخه های جدید وضعیت ش خراب شده و تقریبا کار نمیکنه. چند دقیقه کار میکنه و قطع میشه
اما نسخه های قبلی مثل 1.9.5 یا 1.9.10 بهتر کار میکنه
اما همشون مشکل سرعت دارن و خیلی اذیت کننده هست.
همین جا افراد مختلفی گزارش سرعت پایین این روش رو دادن. لازم هست معماری اسکریپت تغییر کنه تا بهتر بشه. من یک سری پیشنهاد دادم به برنامه نویس پروژه امیدوارم ترتیب اثر بده
#1147

therealaleph:
Reviewed via Anthropic Claude.

@hfjksgk — short answer: Full mode is **structurally slower** than apps_script mode. Don't buy a VPS for raw speed; buy one for:
1. Sites apps_script can't reach (chatgpt.com, claude.ai, x.com login, full Discord/Telegram).
2. Long-lived TCP / WebSocket connections.
3. UDP traffic.

[...]

Issue #1106, opened 2026-05-12
Full mode is extremely slow, like not working at all from yesterday


NightmareYeager:
this my log on android device: 

2026-05-12T16:55:27.092500Z  INFO Loaded MITM CA from /data/user/0/com.therealaleph.mhrv/files/ca/ca.crt
2026-05-12T16:55:27.092717Z  INFO tunnel mux: 4 deployment(s), 30 concurrent per deployment
2026-05-12T16:55:27.092758Z  INFO batch coalesce: step=309ms max=520ms
2026-05-12T16:55:27.093018Z  WARN Listening HTTP   on 0.0.0.0:8080 — set your browser HTTP proxy to this address.
2026-05-12T16:55:27.093040Z  WARN Listening SOCKS5 on 0.0.0.0:1081 — xray / Telegram / app-level SOCKS5 clients use this.
2026-05-12T16:55:27.115100Z  INFO runTun2proxy: cli=tun2proxy --proxy socks5://127.0.0.1:1081 --tun-fd 126 --dns virtual --verbosity info --close-fd-on-drop true --udpgw-server 198.18.0.1:7300
2026-05-12T16:55:27.213154Z  INFO SOCKS5 CONNECT -> 216.239.34.223:443
2026-05-12T16:55:27.213290Z  INFO dispatch 216.239.34.223:443 -> full tunnel (via batch mux)
2026-05-12T16:55:27.216827Z  INFO SOCKS5 CONNECT -> 142.250.21.138:443
2026-05-12T16:55:27.216894Z  INFO dispatch 142.250.21.138:443 -> full tunnel (via batch mux)
2026-05-12T16:55:27.218342Z  INFO SOCKS5 CONNECT -> 142.250.23.95:443
2026-05-12T16:55:27.218454Z  INFO dispatch 142.250.23.95:443 -> full tunnel (via batch mux)
2026-05-12T16:55:27.221495Z  INFO SOCKS5 CONNECT -> 142.251.23.139:443
2026-05-12T16:55:27.221589Z  INFO dispatch 142.251.23.139:443 -> full tunnel (via batch mux)
2026-05-12T16:55:27.221940Z  INFO SOCKS5 CONNECT -> 142.250.23.95:443
2026-05-12T16:55:27.221997Z  INFO dispatch 142.250.23.95:443 -> full tunnel (via batch mux)
[...]

PR #1108, opened 2026-05-12
fix(tunnel): batch header read honors request_timeout_secs (#1088)

### LINKED ISSUE #1088: Slow download speeds for small files
I've noticed that the download speed varies significantly depending on file size:

Platform: Android
App Version: Latest version and also previous versions
Mode: Full Tunnel

• Large files (>10 MB): Download at a decent speed
• Small files (<10 MB): Download speed is capped around 50-80 KB/s

I tested this in Telegram as well:
• When downloading a big file, it starts by downloading a 512 KB chunk and then continues with a decent speed
• For smaller files, it starts with a 32 KB chunk and the speed is limited to 50-80 KB/s

However, if I simultaneously start downloading multiple small files, the speed multiplies for each file:
• 2 files (5 MB each): speed increases to 100-160 KB/s
• 4 files: speed increases to ~300 KB/s

Is there a way to fix this so that smaller files also download quickly?

I'm using full tunnel mode.

Extra note: Using the download speed boost option in unofficial Telegram clients improves the download speed for both small and large files. However, there's still a difference between the download speed of small and large files.

Logs attached:

• 1mb.txt — log for 1 MB worth of video from a 6 MB video in Telegram (download speed boost was on, speed ~50 KB/s)
• 13.9mb.txt — log for a 13.9 MB video in Telegram (download speed boost was on, speed ~300-600 KB/s)

One more note:

I believe this behavior may not be universal across all apps/websites. For example:
• Downloading 6 MB from Cloudflare's speedtest — speed was good
• Downloading from Instagram — the slow-down behavior can be observed. in fact, all videos with all sizes load slowly. but in the web version the behavior is different and needs further checks
edit: the web version of Instagram is fine. i guess the web version of telegram should be fine too.

dazzling-no-more:
## Problem

Issue #1088: under Full mode, a single slow Apps Script edge cascade-killed every in-flight tunnel session sharing its batch. Users on 1.9.21+ saw frequent 10 s "batch timeout" errors and lost download progress on Telegram / browser sessions.

Root cause: `read_http_response` in `domain_fronter.rs` had a hardcoded 10 s header-read timeout that ran **inside** `tunnel_batch_request_to` — independent of and shorter than the outer `tokio::time::timeout(batch_timeout, ...)` in `fire_batch`. Apps Script cold starts routinely land in the 8-12 s range (PR #1040's A/B test recorded 4/30 H1 batches timing out at exactly 10 s after the H2→H1 switch), so the inner cliff was firing as a false-positive batch timeout well before `request_timeout_secs` (default 30 s) could.

Secondary issue: even with a parameterized timeout, the per-read `timeout(d, stream.read(...))` form would silently extend its budget if a peer drip-fed bytes just under `d` each. A slow edge could keep the loop alive past the outer `batch_timeout` and defeat the whole wiring.

## Fix

Two changes, both in `domain_fronter.rs`:

1. **`tunnel_batch_request_to` passes `batch_timeout` to the header read** via the new `read_http_response_with_header_timeout` helper. `Config::request_timeout_secs` is now the only knob controlling how long we wait for an Apps Script edge to start responding. Other callers (relay path, exit-node) keep the historical 10 s value.

2. **Header read uses a single absolute deadline** (`tokio::time::timeout_at(deadline, ...)`) instead of per-read `timeout()`. Total elapsed across all header reads is bounded by `header_read_timeout`, regardless of read cadence.

Bonus in `tunnel_client.rs`:

3. **`TunnelMux::reply_timeout` co-varies with `batch_timeout`**: computed at construction as `fronter.batch_timeout() + 5 s slack` instead of the fixed 35 s const. Operators raising `request_timeout_secs` no longer have sessions abandon `reply_rx` just before `fire_batch`'s HTTP round-trip would complete.

## What this does NOT do

Earlier iterations of this branch added client-side batch retry (re-fire the same payload against a different deployment on timeout). That code was removed before merge: tunnel-node's `drain_now` mutates the per-session buffer when building a poll response, so a lost response means lost bytes — replaying an empty poll silently skips bytes (gap on the client side), and replaying a payload op double-writes (corrupts TCP / double-delivers UDP). No op shape is safe to replay without server-side ack / sequence support. Sessions whose batch times out re-poll on the next tick, same recovery surface as pre-#1088.

## Changes

- `domain_fronter.rs`:
  - `tunnel_batch_request_to`: pass `batch_timeout` to header read (+9 lines).
  - Add `read_http_response_with_header_timeout(stream, deadline)` (+25 lines).
  - Replace per-iteration `timeout()` with `timeout_at(deadline, ...)` in the header loop.
- `tunnel_client.rs`:
  - Add `TunnelMux::reply_timeout` field, derived from `batch_timeout` at construction.
  - `tunnel_loop` reads `mux.reply_timeout()` instead of the `REPLY_TIMEOUT` const.
  - Replace `REPLY_TIMEOUT: Duration` const with `REPLY_TIMEOUT_SLACK: Duration = 5 s`.

## Test coverage

Three new tests, all using `tokio::time::pause()` so they run in < 1 s of wall time:

- `read_http_response_respects_configured_header_timeout` — response starts at virtual T=15 s; must succeed under a 30 s budget. Catches any regression to the hardcoded 10 s inner cliff.
- `read_http_response_header_deadline_is_total_not_per_read` — drip-feeds 38 header bytes at 3 s/byte. Each read returns within 3 s (< 10 s per-read budget) but total exceeds 10 s. Asserts `FronterError::Timeout`. Catches any regression to per-iteration `timeout()`.
- `mux_reply_timeout_tracks_batch_timeout_plus_slack` — constructs a real `DomainFronter` with `request_timeout_secs: 60` via `TunnelMux::start`; asserts `mux.reply_timeout() == 60 s + REPLY_TIMEOUT_SLACK`. Catches any hardcoded-value regression in the runtime derivation.

## Verification

- `cargo test`: 231/231 ✅ (was 228 pre-patch + 3 new)
- `cargo check`: clean ✅
- `cargo clippy -W unused`: clean ✅

## Interaction with PR #1040 / #1041 (H2 → H1 switch)

PR #1040's A/B numbers showed 4/30 H1 batches timing out at 10 s after the H2→H1 switch — that was this same 10 s inner cliff. This PR addresses the residual timeout regression #1040 introduced without reverting the H2→H1 decision, which remains correct for tunnel-batch coalesced ops.

Closes #1088.

### DIFF
--- a/src/domain_fronter.rs
+++ b/src/domain_fronter.rs
@@ -3125,8 +3125,16 @@ impl DomainFronter {
         entry.stream.write_all(&payload).await?;
         entry.stream.flush().await?;
 
+        // Use the configured `request_timeout_secs` for the header read,
+        // not the hardcoded 10 s default. With Apps Script cold starts
+        // routinely landing in the 8–12 s range, the 10 s cliff was
+        // firing as a false-positive batch timeout (issue #1088), killing
+        // every in-flight tunnel session under it. The outer
+        // `tokio::time::timeout(batch_timeout, ...)` in `fire_batch`
+        // remains the authoritative bound on total batch round-trip time.
+        let batch_timeout = self.batch_timeout();
         let (mut status, mut resp_headers, mut resp_body) =
-            read_http_response(&mut entry.stream).await?;
+            read_http_response_with_header_timeout(&mut entry.stream, batch_timeout).await?;
 
         // Follow redirect chain
         for _ in 0..5 {
@@ -3139,7 +3147,8 @@ impl DomainFronter {
             );
             entry.stream.write_all(req.as_bytes()).await?;
             entry.stream.flush().await?;
-            let (s, h, b) = read_http_response(&mut entry.stream).await?;
+            let (s, h, b) =
+                read_http_response_with_header_timeout(&mut entry.stream, batch_timeout).await?;
             status = s; resp_headers = h; resp_body = b;
         }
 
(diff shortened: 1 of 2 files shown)


dazzling-no-more:
Confirmed by a fresh field report in #1106 — log shows `rtt=10.301966975s` followed by `batch failed: timeout` and a cascade of `connect_data error: timeout` across every session in the batch, on a 4-deployment config. Exact signature of the hardcoded 10s header-read cliff this PR removes.

Closes #1088, #1106, #1107.

PR #1128, opened 2026-05-13
Code.gs

opsa3198-a11y:


therealaleph:

Declining — this PR replaces the placeholder `AUTH_KEY = "CHANGE_ME_TO_A_STRONG_SECRET"` in `Code.gs` with a fixed string (`"plmnbvcxzaq"`).

The placeholder exists specifically so that `_doSingle` / `_doBatch` reject requests when it hasn't been changed — see the early-return:

```js
if (AUTH_KEY === "CHANGE_ME_TO_A_STRONG_SECRET") {
  return _json({ e: "configure AUTH_KEY in Code.gs" });
}
```

Hardcoding a real value in the upstream file would:
1. Break the fail-closed safety net for every fresh deployment
2. Bake your own secret into every user's deployment

Every user must set their **own** `AUTH_KEY` to a strong secret they generate themselves, then copy the matching value into `mhrv-rs` `config.json` as `auth_key`. The README + Code.gs comments cover this.

If you intended to share a setup tip (e.g. about character classes some users have trouble with — see #1131 BuffOvrFlw's note about "ASCII-only auth keys work better than password-manager-generated ones"), open an issue instead, not a PR mutating the placeholder.

Same pattern as past declines (#693, #531, #563). Not planned.

Issue #959, opened 2026-05-09
using firefox i can only connect to sites like google and twitter using incognito mode


Sherrybmd:
instant unable to connect error, further explanation leads to "you are not securely connected to this site.
youtube in incognito mode loads but videos instantly fail to play after clicking on play. 
linux mint, works fine on android using chrome, same issue in windows firefox but windows doesnt need incognito, youtube just fails to play.

log:
>>>
2026-05-09T10:29:56.996015Z  INFO Both the glow and wgpu renderers are available. Using glow.
2026-05-09T10:29:57.090606Z  INFO Guessed window scale factor: 1
2026-05-09T10:29:58.934724193Z  [ui] starting proxy...
2026-05-09T10:29:58.935325Z  INFO Loaded MITM CA from /home/sherry/.config/mhrv-rs/ca/ca.crt
2026-05-09T10:29:58.935474009Z  [ui] listening HTTP 127.0.0.1:8085 SOCKS5 127.0.0.1:8086
2026-05-09T10:29:58.935606Z  WARN Listening HTTP   on 127.0.0.1:8085 — set your browser HTTP proxy to this address.
2026-05-09T10:29:58.935633Z  WARN Listening SOCKS5 on 127.0.0.1:8086 — xray / Telegram / app-level SOCKS5 clients use this.
2026-05-09T10:29:59.156367Z  INFO h2 connection established to relay edge
2026-05-09T10:30:00.330981Z  INFO h2 fast path active; h1 fallback pool pre-warmed with 2 connection(s)
2026-05-09T10:30:10.280805398Z  [ui] running test...
2026-05-09T10:30:10.281055Z  INFO test: probing https://api.ipify.org/?format=json via SNI=www.google.com @ 216.239.38.120
2026-05-09T10:30:10.545399Z  INFO h2 connection established to relay edge
2026-05-09T10:30:10.829679Z  INFO dispatch waa-pa.clients6.google.com:443 -> sni-rewrite tunnel (Google edge direct)
2026-05-09T10:30:10.829747Z  INFO SNI-rewrite tunnel -> waa-pa.clients6.google.com:443 via 216.239.38.120 (outbound SNI=www.google.com)
2026-05-09T10:30:12.506329Z  INFO test: PASS: end-to-end verified (response IP = 5.126.187.83).
2026-05-09T10:30:12.506371147Z  [ui] test result: pass
[...]

PR #1159, opened 2026-05-13
fix(cert): install MITM CA into LibreWolf NSS stores

### LINKED ISSUE #1145: bing any popular site
bing.com is most likely a safe site, but a secure connection could not be established. This issue is caused by MasterHttpRelayVPN, which is either software on your computer or your network.

What can you do about it?

bing.com has a security policy called HTTP Strict Transport Security (HSTS), which means that LibreWolf can only connect to it securely. You can’t add an exception to visit this site.

dazzling-no-more:
## Problem

Issue #1145: LibreWolf users get `MOZILLA_PKIX_ERROR_MITM_DETECTED` when visiting HSTS-protected sites (bing.com, youtube.com, ...) through MasterHttpRelayVPN's MITM mode. HSTS gives no "Add Exception" affordance, so the user is fully locked out of those sites despite the OS-level CA install having succeeded.

Root cause: `cert_installer.rs` only scanned Firefox profile roots (`~/.mozilla/firefox`, the snap variant, `%APPDATA%\Mozilla\Firefox\Profiles`, `~/Library/Application Support/Firefox/Profiles`). LibreWolf is a Firefox fork with strict privacy defaults; it shares Firefox's NSS DB layout and respects the same `security.enterprise_roots.enabled` pref, but stores its profile tree under its own app dir. Neither the per-profile `certutil -A` install nor the `user.js` enterprise-roots auto-trust fallback ever touched LibreWolf, so the browser never trusted our CA.

Same failure mode behind already-closed #955 and #959 (Firefox-fork users reporting the identical "secure connection could not be established" symptom).

## Fix

Extend Mozilla-family profile discovery to cover LibreWolf on every supported platform. No behavioural change for Firefox installs.

### Changes in `src/cert_installer.rs`

- **Renamed** `firefox_profile_dirs()` → `mozilla_family_profile_dirs()`. Same flat-vec return type so all five call sites read identically; the rename is signposting only.
- **Extracted `mozilla_family_profile_roots(os, home, appdata, xdg_config_home)`**: pure helper returning the per-OS root list. Takes env values as args so unit tests pin the layout without env mutation or filesystem touches.
- **Extracted `discover_profile_dirs(roots)`**: pure FS-walker that filters each root's children to those containing `cert9.db` (NSS sql:) or `cert8.db` (legacy dbm:). Splitting it out lets a tempdir test cover the filter directly instead of indirectly through env-coupled callers.
- **Added LibreWolf roots**:
  - macOS: `~/Library/Application Support/LibreWolf/Profiles`
  - Windows: `%APPDATA%\LibreWolf\Profiles`
  - Linux: four layouts, because LibreWolf migrated mid-project:
    - `~/.librewolf` — legacy Firefox-style layout, still present on older installs.
    - `${XDG_CONFIG_HOME:-~/.config}/librewolf/librewolf` — current XDG layout (empty `XDG_CONFIG_HOME` is treated as unset per the XDG Base Directory spec).
    - Both of the above again under `~/.var/app/io.gitlab.librewolf-community/` for the Flatpak sandbox, which redirects HOME inside the container.
- **Renamed** `enable_firefox_enterprise_roots` → `enable_mozilla_enterprise_roots` (and `disable_*` likewise), since the pref now gets written for LibreWolf profiles too. Three internal log strings (`firefox profile {} ...`) updated to `mozilla profile {}` so debug output reflects the broader scope.
- **User-facing log updates** in 4 paths (NSS install warn, certutil-missing debug, removal warn, partial-cleanup warn) now mention `Firefox/LibreWolf/Chrome` instead of `Firefox/Chrome`, so LibreWolf users see actionable guidance when something needs manual cleanup.

### Non-changes (deliberate)

- **Floorp / Waterfox / Mullvad Browser** — same Firefox-fork pattern, same fix shape, but no open issues from those user populations. Held off to keep this PR scoped to the reproducible bug. Easy follow-up if/when reported.
- **macOS `~/.librewolf` and Windows `%USERPROFILE%\.librewolf`** — both appear in the LibreWolf docs but only as `librewolf.overrides.cfg` (the autoconfig file) locations, not as profile-data locations containing `cert9.db`. Including them would mean scanning paths we have no evidence are profile roots. If users report MITM_DETECTED on those layouts, a one-line addition + test covers it.

## Test plan

- [x] `cargo test --lib` — 239 passing (was 236 pre-patch + 8 new in `cert_installer`).
- [x] `cargo build` — clean.
- [x] `rustfmt --check src/cert_installer.rs` — clean.
- [x] New unit tests pin every supported layout — Linux (firefox + snap + librewolf legacy + XDG default + XDG override + Flatpak ×2), macOS, Windows, plus edge cases (empty `XDG_CONFIG_HOME` falls back to default; missing `%APPDATA%` no-ops; unknown OS yields empty).
- [x] New tempdir test for `discover_profile_dirs` constructs `abc.default/cert9.db`, `legacy.profile/cert8.db`, a sibling dir without a DB, and a stray `profiles.ini` file plus a non-existent root — asserts only the two real profiles come back.
- [ ] Manual repro on a Linux box with LibreWolf installed: install CA via `mhrv-rs-ui`, restart LibreWolf, visit https://bing.com — expect the page to load instead of `MOZILLA_PKIX_ERROR_MITM_DETECTED`. (Recommend issue reporter @persiansunny re-run the original repro from #1145 to confirm.)

Closes #1145. Likely also addresses #955 and #959 (same root cause class), though those users would need to confirm.

### DIFF
--- a/src/cert_installer.rs
+++ b/src/cert_installer.rs
@@ -54,7 +54,7 @@ impl RemovalOutcome {
 /// When running as root via `sudo`, the process's `HOME` / `USER`
 /// environment reflects **root**, not the user who invoked the command.
 /// That breaks every user-scoped cert path this module touches —
-/// `data_dir()` resolves to root's config dir, `firefox_profile_dirs()`
+/// `data_dir()` resolves to root's config dir, `mozilla_family_profile_dirs()`
 /// scans root's profiles, macOS `login.keychain-db` is root's. The
 /// removal then operates on paths that probably don't exist, reports
 /// success, and leaves the real user's CA trusted.
@@ -840,10 +840,10 @@ fn remove_windows() {
     }
 }
 
-// ---------- NSS (Firefox + Chrome/Chromium on Linux) ----------
+// ---------- NSS (Firefox + LibreWolf + Chrome/Chromium on Linux) ----------
 
 /// Best-effort install of the CA into all discovered NSS stores:
-///   1. Every Firefox profile (each has its own cert9.db).
+///   1. Every Firefox/LibreWolf profile (each has its own cert9.db).
 ///   2. On Linux, the shared Chrome/Chromium NSS DB at ~/.pki/nssdb —
 ///      this is the one update-ca-certificates does NOT populate, and
 ///      missing it was the real blocker for Chrome users who'd installed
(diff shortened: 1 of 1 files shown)


BuffOvrFlw:
Would you also please add support for GNU IceCat?
the directory is:
~/.mozilla/icecat/

dazzling-no-more:
> the directory is:
> ~/.mozilla/icecat/

Done, please test and report back.

Ardeshir0506 (approved):

PR #1201, opened 2026-05-14
exit_node.ts

opsa3198-a11y:


therealaleph:
Declined — same author / same string / same pattern as your already-declined PR [#1128](https://github.com/therealaleph/MasterHttpRelayVPN-RUST/pull/1128) (which I closed yesterday with explanation). You're hardcoding the same fixed PSK value (`"plmnbvcxzaq"`) into the placeholder `CHANGE_ME_TO_A_STRONG_SECRET` — this time targeting `exit_node.ts` instead of `Code.gs`.

The placeholder values across these files (`Code.gs`, `exit_node.ts`, `CodeFull.gs`, `Code.cfw.gs`) exist as **fail-closed safety nets** — every consumer of them has explicit `if (PSK === "CHANGE_ME_...")` guards that refuse to run until the user replaces it with their own secret. Hardcoding a real value would:

1. Break the fail-closed safety net for every fresh deployment
2. Bake a known secret into every user's deployment
3. Effectively give anyone who knows `"plmnbvcxzaq"` access to every exit-node deployed from this repo

Pattern reading: this is the same shape as past supply-chain attempts (#693, #531, #563, #1128). Each one targeting a different config-secret placeholder, each one hardcoding a fixed value.

I'm declining this PR and will decline any future PRs from this account that touch placeholder secrets. If you have a genuine improvement to suggest — like a stronger default PSK *format* (different length, different character classes, instructions to generate with `openssl rand -hex 32`) — open an issue or PR with that framing instead of substituting a hardcoded value.

Not planned. Locking would be excessive — leaving open for the project record.

Issue #1047, opened 2026-05-11
502 error (Script Completed but did not return anything) on downloading files.


Aaplusthestar:
using version 1.9.21 and windows 10 as the OS with MOKHABERAT as the ISP.
Everything else works normally (exepct anti-bot websites), here's the log:
2026-05-11T05:16:43.183503Z  INFO dispatch shc-dist.lostsig.co:443 -> MITM + Apps Script relay (TLS detected)
2026-05-11T05:16:43.183526Z  INFO MITM TLS -> shc-dist.lostsig.co:443 (socks_host=shc-dist.lostsig.co, sni=shc-dist.lostsig.co)
2026-05-11T05:16:43.188045Z  INFO relay GET https://shc-dist.lostsig.co/downloads/compo2013/[SHC2013]%20[compo2013]%203%20-%20Metal%20Sonic%20Hyperdrive%20-%20By%20Kaz.bin
2026-05-11T05:16:46.471406Z ERROR Relay failed: bad response: no json in: <!DOCTYPE html><html><head><title>Web App</title></head><body><p>The script completed but did not return anything.</p></body></html>
2026-05-11T05:16:58.381306Z  INFO stats: relay=168 (1447KB) failures=2 coalesced=0 cache=0/0 (0% hit, 0KB) scripts=5/5 active h2-success=170/170 (100%)
2026-05-11T05:17:04.186338Z ERROR Relay failed: bad response: no json in: <!DOCTYPE html><html><head><title>Web App</title></head><body><p>The script completed but did not return anything.</p></body></html>
this happens on this specfic website. every other thing i wanted to download with was successful other than this website (sonichacking.org) on chrome

Issue #1019, opened 2026-05-10
Error: no json in


Official100Ra:
Hello, there is a problem
When I run the program, I encounter this Error

`ERROR Relay failed: bad response: no json in: <!DOCTYPE html><html><head><title>Web App</title></head><body><p>The script completed but did not return anything.</p></body></html>`

The program is in Apps Script mode.

therealaleph:
Reviewed via Anthropic Claude.

@Official100Ra — `The script completed but did not return anything` is the **default Apps Script Web App template body** — Google's placeholder when the deployed script doesn't return a  response.

This means the Code.gs content was not pasted into the script editor before deploying. Same root cause as #973 / #948 / #755.

Fix:
1. Open https://script.google.com → your project.
2. The default file is `Code.gs` with one empty `myFunction()`. Replace its **entire** contents with the file at [`assets/apps_script/Code.gs`](https://github.com/therealaleph/MasterHttpRelayVPN-RUST/blob/main/assets/apps_script/Code.gs) (Ctrl+A → Ctrl+V).
3. Set the constant at the top of the file:
   ```
   const AUTH_KEY = "<your secret>";  // must match auth_key in mhrv-rs config.json
   ```
4. Save (floppy icon).
5. **Deploy → New deployment** (NOT "Test deployment", NOT "Edit existing"):
   - Type: Web app
   - Execute as: Me (your account)
   - Who has access: **Anyone, even anonymous** ← critical
6. Copy the new deployment ID (the long `AKfycb...` string from the URL).
7. Update `script_id` in mhrv-rs `config.json` to the new deployment ID.
8. Restart mhrv-rs.

Verify: in browser, visit `https://script.google.com/macros/s/<your-id>/exec` directly — should return JSON `{"e":"method_not_allowed"}` (because GET hits Code.gs's POST-only handler), NOT the default `<!DOCTYPE html>...` placeholder.

Closing as duplicate of #973. Reopen with the result of the direct-browser GET probe if the placeholder persists.

Issue #973, opened 2026-05-09
VPS NOT WORKING !


ehsan272727:
I ran these after connecting with ssh

curl -fsSL https://get.docker.com | sh

docker run -d \
  --name mhrv-tunnel \
  --restart unless-stopped \
  -p 8080:8080 \
  -e TUNNEL_AUTH_KEY="$SECRET" \
  ghcr.io/therealaleph/mhrv-tunnel-node:latest

but Full tunnel mode does not work at all

dazzling-no-more:
Did you actually set the secret? Also you need to deploy the script with the same secret.

ehsan272727:
I did, but it doesn't work

ehsan272727:
2026-05-10T03:53:32.884773Z ERROR batch failed (script AKfycbzC): got the v1.8.0 decoy/placeholder body — could be (1) AUTH_KEY mismatch between mhrv-rs config and Code.gs (run a direct curl probe against the deployment to verify), (2) Apps Script execution timeout or per-100s quota tear (try lowering parallel_concurrency in config), (3) Apps Script internal hiccup (transient, retry next batch), or (4) ISP-side response truncation (#313 pattern, try a different google_ip). To distinguish (1) from the rest: set DIAGNOSTIC_MODE=true at the top of Code.gs + redeploy as new version — only AUTH_KEY mismatch returns this body in diagnostic mode.
2026-05-10T03:53:32.884797Z ERROR tunnel connect_data error for www.youtube.com:443: bad response: no json in batch response: <!DOCTYPE html><html><head><title>Web App</title></head><body><p>The script completed but did not return anything.</p></body></html>
2026-05-10T03:53:32.884799Z ERROR tunnel connect_data error for www.youtube.com:443: bad response: no json in batch response: <!DOCTYPE html><html><head><title>Web App</title></head><body><p>The script completed but did not return anything.</p></body></html>

therealaleph:
Reviewed via Anthropic Claude.

@ehsan272727 — `The script completed but did not return anything` is the **default Apps Script Web App template body** — Google's placeholder when the deployed script doesn't return a `ContentService.createTextOutput(...)` response.

This means: **`CodeFull.gs` was not pasted into the script editor before deploying** — Google deployed an empty template instead.

Fix:
1. https://script.google.com → open your project.
[...]

Issue #948, opened 2026-05-09
مشکل از کجاست؟


monjinjast:
2026-05-09T06:36:49.195998Z ERROR Relay failed: bad response: no json in: <!DOCTYPE html><html><head><title>Web App</title></head><body><p>The script completed but did not return anything.</p></body></html>

therealaleph:
Same root cause as #973 + #755 — Apps Script Web App returning the default 'The script completed but did not return anything' template means `Code.gs`/`CodeFull.gs` content was not pasted into the script editor before deploying. See [#973](https://github.com/therealaleph/MasterHttpRelayVPN-RUST/issues/973) for full fix steps. Closing as duplicate.

PR #1209, opened 2026-05-14
fix(exit-node): fix double-wrapped response envelope and panic in error paths

### LINKED ISSUE #1022: mhrv-rs fails decoding exit-node response
I set up exit node and can confirm it works correctly. Direct and no reverse proxy in between.

From a machine with Internet access:

`curl -v -X POST http://[REDACTED]:8445   -H "Content-Type: application/json"   -d '{"k":"REDACTED","u":"https://api.ipify.org/?format=json","m":"GET","h":{}}'
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying [REDACTED]:8445...
* Established connection to [REDACTED] ([REDACTED] port 8445) from 192.168.137.45 port 41016 
* using HTTP/1.x
> POST / HTTP/1.1
> Host: [REDACTED]:8445
> User-Agent: curl/8.20.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 130
> 
* upload completely sent off: 130 bytes
< HTTP/1.1 200 OK
< content-type: application/json
< vary: Accept-Encoding
< content-length: 225
< date: Sun, 10 May 2026 15:54:05 GMT
< 
* Connection #0 to host [REDACTED]:8445 left intact
{"s":200,"h":{"cf-cache-status":"DYNAMIC","cf-ray":"9f9a1edaba281953-FRA","content-type":"application/json","date":"Sun, 10 May 2026 15:54:05 GMT","server":"cloudflare","vary":"Origin"},"b":"eyJpcCI6Ijc3LjIyMy4yMTUuNDUifQ=="}[~:9]
`

MHRV-RS exit node config fragment:

`   "exit_node": {
    "_comment": "Master switch. Set false to disable exit-node entirely without removing the config. Default false.",
    "enabled": true,
    "_comment_relay_url": "Public URL of your deployed exit-node handler (assets/exit_node/exit_node.ts running on Deno Deploy, fly.io, your own VPS, etc.).",
    "relay_url": "http://[REDACTED]:8445",
    "_comment_psk": "Pre-shared key — must match the PSK constant in your deployed source. Generate with: openssl rand -hex 32",
    "psk": "[REDACTED]",
    "_comment_mode": "selective: only `hosts` route via exit node (recommended). full: every request routes via exit node (slower, ~250-500ms extra hop).",
    "mode": "full",
    "_comment_hosts": "Hostnames to route through the exit node. Matches exact OR dot-anchored suffix (chatgpt.com covers api.chatgpt.com etc.). Extend for any CF-anti-bot blocked sites you need.",
    "hosts": [
      "chatgpt.com",
      "claude.ai",
      "x.com",
      "grok.com",
      "openai.com",
      "aistudio.google.com",
      "ai.google.dev"
    ]
}`

From the rust logs:
`2026-05-10T15:58:48.703406Z  WARN test: 200 OK without ipify 'ip' field — upstream may be broken. body: {"s":200,"h":{"cf-cache-status":"DYNAMIC","cf-ray":"9f9a25a6de368e2c-FRA","content-type":"application/json","date":"Sun, 10 May 2026 15:58:44 GMT","server":"cloudflare","vary":"Origin"},"b":"eyJpcCI6I
2026-05-10T15:58:48.703439859Z  [ui] test result: fail
2026-05-10T15:58:48.703443Z  INFO fetch_ips_from_api disabled, using static fallback
2026-05-10T15:58:48.703447Z DEBUG Connection{peer=Client}: send frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(0) }
2026-05-10T15:58:48.703454Z DEBUG Connection{peer=Client}: Connection::poll; connection error error=GoAway(b"", NO_ERROR, Library)`

Test always fails with vps exit-node. As you can see the response line has base64 response (did the logger truncate it?) and decoding the response manually shows the correct response with ip field. The rust code fails to decode base64?

CaptainMirage:
Fixed two bugs in the exit-node relay path.

**Commit 1** fixes #1022 — Code.gs was double-wrapping the exit node's
{s,h,b} response envelope, causing browsers to receive raw JSON instead
of page content. Adds the raw-return mode (req.r === true) to _doSingle
and fixes parse_exit_node_response to handle HTTP framing prefixes and
stale content-encoding headers.

**Commit 2** fixes a panic (SIGILL/core dump) in four error format strings
that sliced a &str at byte offset 200, which crashes when the response
contains multi-byte UTF-8 characters (quota error pages, binary responses).
Replaced with char-aware truncation.

Tested against claude.ai, chatgpt.com, openai.com, perplexity.ai and general browsing with both
WARP and direct exit-node modes.

### DIFF
--- a/assets/apps_script/Code.gs
+++ b/assets/apps_script/Code.gs
@@ -161,6 +161,9 @@ function _doSingle(req) {
     return _json({ e: "bad url" });
   }
 
+  // ── Optional cache path ────────────────────────────────
+  // Only entered when CACHE_SPREADSHEET_ID is configured and
+  // the request qualifies as a public, cachable GET.
   if (_canUseCache(req)) {
     var cached = _getFromCache(req.u, req.h);
     if (cached) {
@@ -181,25 +184,41 @@ function _doSingle(req) {
         cached: false,
       });
     }
-    // _fetchAndCache returned null → fall through to normal relay
+    // If _fetchAndCache returns null (spreadsheet unavailable),
+    // fall through to the normal relay path below.
   }
 
-  var opts = _buildOpts(req);
-  var resp = UrlFetchApp.fetch(req.u, opts);
+  // ── Normal relay (cache disabled or unavailable) ────────
+  // Wrap the fetch + body encode in try/catch so any failure surfaces as
+  // a JSON error envelope the Rust client can parse. Without this, throws
+  // from UrlFetchApp.fetch (URL too long, payload too large, quota
+  // exhausted, 6-minute execution timeout) or from base64Encode (response
+  // body near Apps Script's ~50 MB ceiling can blow the V8 heap during
+  // encode) propagate unhandled, and Apps Script serves its default
+  // `<title>Web App</title>` HTML error page — which the client then
+  // reports as "Relay failed: bad response: no json in: <title>Web App>..."
+  // and the user has no signal as to the actual cause. Mirrors the
+  // per-item try/catch in _doBatch below.
+  try {
+    var opts = _buildOpts(req);
+    var resp = UrlFetchApp.fetch(req.u, opts);
 
-  // Raw-return mode for exit-node path.
-  // r:true = return destination body verbatim so Rust gets {s,h,b} unwrapped.
-  if (req.r === true) {
-    return ContentService
-      .createTextOutput(resp.getContentText())
-      .setMimeType(ContentService.MimeType.JSON);
+    // Raw-return mode for exit-node path.
+    // r:true = return destination body verbatim so Rust gets {s,h,b} unwrapped.
+    if (req.r === true) {
+      return ContentService
+        .createTextOutput(resp.getContentText())
+        .setMimeType(ContentService.MimeType.JSON);
+    }
+
+    return _json({
+      s: resp.getResponseCode(),
+      h: _respHeaders(resp),
+      b: Utilities.base64Encode(resp.getContent()),
+    });
+  } catch (err) {
+    return _json({ e: "fetch failed: " + String(err) });
   }
-
-  return _json({
-    s: resp.getResponseCode(),
-    h: _respHeaders(resp),
-    b: Utilities.base64Encode(resp.getContent()),
-  });
 }
 
 // ── Batch Request ──────────────────────────────────────────
--- a/assets/apps_script/Code.gs
+++ b/assets/apps_script/Code.gs
@@ -161,9 +161,6 @@ function _doSingle(req) {
     return _json({ e: "bad url" });
   }
 
-  // ── Optional cache path ────────────────────────────────
-  // Only entered when CACHE_SPREADSHEET_ID is configured and
-  // the request qualifies as a public, cachable GET.
   if (_canUseCache(req)) {
     var cached = _getFromCache(req.u, req.h);
     if (cached) {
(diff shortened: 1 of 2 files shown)


therealaleph:

Two real fixes I want to land, mixed with one regression I don't.

**Good (want to keep):**

1. **Char-aware truncation** in 4 error format strings (`parse_exit_node_response` / `parse_relay_json` / two others). Replaces `&text[..text.len().min(200)]` (which panics at byte-boundary cuts through multi-byte UTF-8) with `&text.chars().take(200).collect::<String>()`. Real SIGILL fix when responses contain Persian-localized quota errors or binary payloads. Worth landing on its own.

2. **`parse_exit_node_response` skip past `\r\n\r\n` HTTP-framing prefix** + add `content-encoding` to the stale-headers strip list. Handles the double-wrap case at the parse level without needing Code.gs changes. Belt-and-suspenders fix — also handles cases where the exit-node response comes back through chained relays that add HTTP framing. Worth landing.

3. **`r: true` raw-return mode in `_doSingle`** is forward-compatible (current clients don't send `r: true`, so it's dead-code-ish today). Fine to land; clients can opt-in later.

**Regression (need to fix before merge):**

4. **The try/catch around `UrlFetchApp.fetch(...)` + `Utilities.base64Encode(...)` was removed in `_doSingle`.** That `try/catch` was added in #1047 commit `e2a4be24` specifically to prevent the placeholder `<title>Web App</title>` HTML from leaking to the client when:
   - URL is too long → throws
   - Payload too large → throws
   - Quota exhausted → throws  
   - Execution timeout (6 min) → throws
   - Base64 of 50MB response blows V8 heap → throws

Without the try/catch, those throws propagate unhandled, Apps Script serves the default HTML error page, and the client surfaces `bad response: no json in: <title>Web App>...` instead of a structured `{ e: "fetch failed: ..." }`. That regresses on every #973 / #1019 / #948 / #755-class issue.

**Diagnostic logging (need to remove before merge):**

5. **`EXIT_DIAG app_body len=...` warnings** added in `domain_fronter.rs::relay_exit_node`. Useful for debugging during the PR work, but they'd flood every user's `RUST_LOG=info` console with full first-200-bytes hexdumps in production. Strip before merge.

**Suggested action:**

If you'd like, I can split this PR's changes — keep (1), (2), (3) and drop (4)'s try/catch removal + (5) the EXIT_DIAG warnings. Or you can do it on your branch:

- Restore the try/catch in `_doSingle` around `UrlFetchApp.fetch(opts)` + `Utilities.base64Encode(resp.getContent())` (just add it back around the new code, keep the `r:true` branch outside the catch).
- Strip the 2 `tracing::warn!("EXIT_DIAG ...")` calls from `relay_exit_node`.

After that, this lands.

Re: #1022 root cause — agree it was double-wrapping. The parse-side fix in (2) is the right architectural shape because it doesn't require a Code.gs/client coordinated upgrade.

---

CaptainMirage:
i kinda didnt realized i removed that i was behind the main branch lol sorry, restored that one and removed the debug stuff, tho i would say it wouldnt clutter the info thing since i did put it in debug trace not warn traces but still not that useful outside of this fix so removed it anyway

therealaleph (changes requested):
Thanks for the fast follow-up here. The two underlying fixes make sense to me, especially the char-aware truncation in the error paths.

One blocker before I can merge this: the current diff still contains the temporary `EXIT_DIAG` block in `src/domain_fronter.rs`:

```rust
tracing::debug!(
    "EXIT_DIAG app_body len={} first_200={:?}",
    app_body.len(),
    String::from_utf8_lossy(&app_body[..app_body.len().min(200)])
);
```

That can put the first bytes of proxied response content into logs. In the exit-node path that may be real destination content, not just relay metadata, so we should not ship it even at `debug`. The PR body/commit message says this was removed, but the patch still has it.

Please drop that diagnostic block and leave the direct `parse_exit_node_response(&app_body)` call. After that I can rerun local verification and merge if the rest stays clean.

---

CaptainMirage:
somehow i missed that one even tho its a line above, it should be good now

i have no idea how it just updated from warn to debug from the last commit, i did in fact commit with both lines gone, again should be fine now

therealaleph (approved):
Verified locally on top of current `main` and together with #1229.

Checks run:

```text
node --check --input-type=commonjs < assets/apps_script/Code.gs
node --check --input-type=commonjs < assets/apps_script/CodeFull.gs
node --check --input-type=commonjs < assets/apps_script/Code.cfw.gs
cargo test --lib
cargo build --release
```

`cargo test --lib` passed 239/239, release build passed, and the final diff has no `EXIT_DIAG` debug logging left. This fixes the exit-node raw response/double-envelope path and strips stale `content-encoding` from exit-node responses.

---

PR #1229, opened 2026-05-16
fix(domain_fronter): prevent panic when brace positions are inverted in fallback JSON extraction

CaptainMirage:
three fallback sites in `finalize_tunnel_response`, `finalize_batch_response`,
and `parse_relay_json` use `text.find('{')` and `text.rfind('}')` to pull a JSON
object out of a messy response, when the response body is binary garbage,
those bytes can appear in any order, `rfind` might find `'}'` at byte 84 while
find finds `'{'` at byte 157. The slice `&text[start..=end]` then panics with
"begin > end" and a SIGILL core dump

Added a bounds check at all three sites: if `start > end` return a structured
BadResponse error instead of slicing and promptly thanos snapping itself

Reproducible whenever Apps Script returns a binary or non-JSON response body,
which happens under quota pressure, transient errors, or when the exit node
returns unexpected content (sometimes managed to reproduce with huge amounts of ping)


CaptainMirage:
welp i somehow managed to pull from the main repo and did this, hold on im gonna fix it

CaptainMirage:
if you may please ignore those 2 commits of trying to merge into the main branch, i somehow managed to run the pull command from an older command and did `git pull upstream main` and so ii rewired all branch tracking to the main repo and i was running around the problem for an hour, pardon

therealaleph (approved):
Verified locally on top of current `main` and together with #1209.

Checks run:

```text
node --check --input-type=commonjs < assets/apps_script/Code.gs
node --check --input-type=commonjs < assets/apps_script/CodeFull.gs
node --check --input-type=commonjs < assets/apps_script/Code.cfw.gs
cargo test --lib
cargo build --release
```

`cargo test --lib` passed 239/239 and release build passed. The added `start > end` guards are a good defensive fix for malformed fallback JSON extraction and apply cleanly with the exit-node changes.

---

PR #1254, opened 2026-05-17
some performance improvements for CodeFull.gs

brightening-eyes:
perf(apps_script): array-join patterns, Object.assign fallback, slice… copy for _dnsRewriteTxid


therealaleph (approved):
Reviewed and locally checked. The CodeFull.gs changes are narrow performance cleanups and preserve the existing response envelope behavior.

Verification run locally in a clean worktree:

```text
node --check /tmp/CodeFull-1254.js
cargo test --lib
```

I am not merging this while the v1.9.29 release workflow is in flight, but the PR itself looks good from code review.

---

PR #1263, opened 2026-05-17
Update config.fronting-groups.example.json

Shjpr9:
- Some minor tweaks to make Github work

We know that many fronting groups we used to use has been recently blocked. 
This commit contains some changes to make Github work better, However, still some domains (like githubassets.com) must pass through Apps Script.
My tests show that these changes will make some actions like navigating through repositories faster and more smooth. 
I'll try my best to scan and find working ips that are on Fastly edge and will make a PR in the future. 

Thanks in advance

### DIFF
--- a/config.fronting-groups.example.json
+++ b/config.fronting-groups.example.json
@@ -44,8 +44,8 @@
     },
     {
       "name": "fastly",
-      "ip": "151.101.128.223",
-      "sni": "files.pythonhosted.org",
+      "ip": "185.199.109.215",
+      "sni": "github.githubassets.com",
       "domains": [
         "redd.it",
         "reddit.com",
--- a/config.fronting-groups.example.json
+++ b/config.fronting-groups.example.json
@@ -45,7 +45,7 @@
     {
       "name": "fastly",
       "ip": "151.101.128.223",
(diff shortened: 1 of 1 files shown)


therealaleph (changes requested):
Thanks for the scan and the GitHub-focused testing. I cannot merge this shape yet because the PR changes the existing `fastly` group from a Fastly/PyPI edge to `185.199.109.215` + `github.githubassets.com`, while leaving the Reddit/Fastly/PyPI domains in that same group. That can regress non-GitHub hosts that currently rely on the Fastly group.

Please split the GitHub asset/fronting changes into a dedicated GitHub group, or keep the existing `fastly` group intact and only move/add GitHub domains where the SNI/IP pair is known to work. The file is valid JSON (`jq empty` passed), but the grouping semantics need to stay safe for the older domains.

---

therealaleph:
Thanks for sending this. This PR is now superseded by the config TOML migration in PR #1317 and release v1.9.34.

`config.fronting-groups.example.json` has been replaced by `config.fronting-groups.example.toml`, and keeping a JSON-only example update open would point users at the old format. Closing this one to keep the docs/examples clean.

---

PR #1309, opened 2026-05-19
fix(pipeline): idle poll flood, data streak breaks, tuning

yyoyoian-pixel:
## Summary
Fixes the v1.9.28+ pipelining regression where the request counter rapidly increases, Instagram videos fail to load, and 15 deployments is slower than 6 with the old code.

### Tuning changes
- **INFLIGHT_OPTIMIST** 2 (unchanged): optimistic starting depth stays at 2
- **INFLIGHT_ACTIVE** 4 (unchanged): max pipeline depth stays at 4
- **Upload overflow** +4 → +2: tighter cap on extra in-flight upload ops
- **MAX_ELEVATED** 30-per-deployment → 10 total: flat elevation cap across all deployments
- **block_stun** default `true` → `false`: STUN/TURN traffic allowed by default

### Pipeline regression fixes
1. **Escalating keepalive backoff** (20ms→80ms→200ms→500ms→2s) when pipeline drains to zero in-flight. The pre-pipelining serial loop had this; the new loop sent polls with zero delay, flooding idle sessions.
2. **Suppress refill timer at IDLE depth** with consecutive empties — the keepalive path with backoff handles that; the refill timer was scheduling empty polls every ~1s regardless.
3. **Stale empty-poll replies don't break data streaks** — a poll queued before data started flowing returns empty as expected. Previously it incremented consecutive_empty and reset consecutive_data mid-stream, causing premature depth drops that killed video streaming throughput (Instagram).
4. **Reduce upload overflow +4→+2** — upload reads were inflating request count beyond the pipeline depth budget.

### Root cause
The old serial loop (pre-v1.9.28) sent one request at a time with escalating backoff (20ms→80ms→200ms→500ms→30s). Idle sessions cost essentially zero requests. The new pipelined loop had no backoff on empty keepalive polls and kept the refill timer firing every 1s at all depths, generating continuous empty requests across all sessions. With 15 deployments × many sessions × no backoff = quota exhaustion.

## Test plan
- [ ] Verify request counter growth rate is comparable to pre-v1.9.28 on idle sessions
- [ ] Confirm Instagram videos load and stream without stalling
- [ ] Verify 15 deployments performs at least as well as 6 deployments with old code
- [ ] Check pipeline ramps up to depth 4 on sustained data flows
- [ ] Verify elevation cap limits total elevated sessions to 10

### DIFF
--- a/src/bin/ui.rs
+++ b/src/bin/ui.rs
@@ -431,7 +431,7 @@ fn load_form() -> (FormState, Option<String>) {
             youtube_via_relay: false,
             passthrough_hosts: Vec::new(),
             block_quic: true,
-            block_stun: true,
+            block_stun: false,
             disable_padding: false,
             force_http1: false,
             tunnel_doh: true,
@@ -695,8 +695,8 @@ struct ConfigWire<'a> {
     /// emit only when the user has explicitly disabled the block.
     #[serde(skip_serializing_if = "is_true")]
     block_doh: bool,
-    /// Default true. Emit only when the user disables STUN/TURN blocking.
-    #[serde(skip_serializing_if = "is_true")]
+    /// Default false. Emit only when the user enables STUN/TURN blocking.
+    #[serde(skip_serializing_if = "is_false")]
     block_stun: bool,
     #[serde(skip_serializing_if = "Vec::is_empty")]
     fronting_groups: &'a Vec<FrontingGroup>,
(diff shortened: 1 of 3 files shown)


therealaleph (approved):
Verified locally on current head `e1014ef57ff5ab58046c15435afde5f2dc54ede0`.

Passed:

- `cargo test --lib`
- `cargo build --release`
- `cargo build --bin mhrv-rs-ui --release --features ui`
- `JAVA_HOME="/Applications/Android Studio.app/Contents/jbr/Contents/Home" ANDROID_HOME="$HOME/Library/Android/sdk" ./gradlew :app:assembleDebug`

Approved for merge.

---

PR #1322, opened 2026-05-20
fix(pipeline): stop idle keepalive polls, reduce idle requests

yyoyoian-pixel:
## Summary
Stops keepalive polling on truly idle sessions and adds gradual backoff with separate idle tier tracking. Reduces idle Apps Script requests from ~1200/5min to ~35/5min for truly idle sessions.

### How it works
1. **Early backoff** (first 3 empty replies): 20ms → 80ms → 4s → 10s
2. **Stop polling after idle_tier > 3** (~16s of empties): session stops sending keepalive polls entirely, just waits for client data
3. Apps maintain their own heartbeats (MQTT PINGREQ, FCM keepalive, Telegram ping) which trigger client writes — those act as natural polls that pick up server-pushed data

### Separate counters
- `consecutive_empty`: controls pipeline depth management (resets to 0 on data — sessions can ramp up immediately)
- `idle_tier`: controls keepalive delay (halves on server-pushed data instead of full reset — prevents push notifications from yo-yoing the backoff)

### Measured results
| Minute | Requests | Notes |
|---|---|---|
| 0:00-1:00 | 23 | Initial ramp-down |
| 1:00-2:00 | 63 | Sessions settling |
| 2:00+ | 7-50 | Steady state (app heartbeats only) |

| Version | Idle requests/5min |
|---|---|
| v1.9.13 (serial) | ~300 |
| v1.9.30 (pipelined, 2s cap) | ~1200 |
| v1.9.31 (pipelined, 2s cap) | ~1000 |
| **This PR (truly idle)** | **~35** |
| **This PR (background apps active)** | **~250-350** (real traffic, not wasted polls) |

Note: sessions receiving periodic server pushes (Instagram MQTT, Telegram, etc.) never reach idle_tier > 3 because they keep getting data. Those requests are real traffic, not wasted keepalive polls. True idle sessions (no server pushes) drop to zero polls.

### No latency impact
- `tokio::select!` races delay timer against client socket reads
- When app writes data (heartbeat, user action), fires immediately
- Pipeline depth uses `consecutive_empty` (resets to 0) — not affected by idle_tier

## Test plan
- [x] Verified idle requests drop from ~1200 to ~35/5min for truly idle sessions
- [x] Verified background app traffic (Telegram, Instagram) still works
- [ ] Verify active browsing/video works without added latency
- [ ] Verify push notifications still arrive promptly
- [ ] Test with customer's setup (15 deployments, Chrome + Instagram only)

### DIFF
--- a/src/tunnel_client.rs
+++ b/src/tunnel_client.rs
@@ -1616,11 +1616,13 @@ async fn tunnel_loop(
         if inflight.is_empty() && !eof_seen {
             let all_legacy = mux.all_servers_legacy();
 
-            // After enough consecutive empties, stop polling and just
-            // wait for client data. Apps maintain their own heartbeats
-            // (MQTT PINGREQ, FCM keepalive, etc.) which trigger client
-            // writes that send data ops — those act as natural polls.
-            if (idle_tier > 1 || (all_legacy && consecutive_empty > 3)) && !client_closed {
+            // If every deployment is legacy and the session has gone
+            // idle, stop polling and just wait for client data. Apps
+            // maintain their own heartbeats (MQTT PINGREQ, FCM keepalive,
+            // etc.) which trigger client writes that send data ops — those
+            // act as natural polls. Mixed fleets must keep polling so
+            // round-robin can still land on a long-poll-capable peer.
+            if all_legacy && (idle_tier > 1 || consecutive_empty > 3) && !client_closed {
                 read_buf.reserve(65536);
                 match reader.read_buf(&mut read_buf).await {
                     Ok(0) => break,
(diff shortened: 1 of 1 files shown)


therealaleph (approved):
Verified locally after adding the mixed-deployment guard: all-legacy idle sessions can stop polling, but mixed fleets keep emitting empty polls so round-robin can still reach a long-poll-capable peer.

Tests:
- `cargo test --lib tunnel_client::tests::tunnel_loop_keeps_polling_when_only_some_deployments_legacy -- --nocapture`
- `cargo test --lib`
- `cargo build --release`

---

PR #1333, opened 2026-05-21
(feat) quota management, transport security, and system resilience

maybeknott:
Implement quota management, transport security, and system resilience features

**- Dynamic SNI Selection Engine**
Modified 'src/domain_fronter.rs' to include a destination-aware SNI mapping matrix. This engine intercepts the target host before connection establishment and selects a Google SNI hostname that mimics legitimate productivity traffic based on the request type. For example, high-bandwidth media streams (googlevideo.com) are cloaked as 'docs.google.com', while API-heavy traffic is mapped to 'developers.google.com'. Other requests are randomized across a rotation pool of common services (mail, drive, maps) to prevent the emergence of a predictable SNI fingerprint that could be flagged by DPI.

**- Upstream Request Fragmentation (Reverse Chunking)**
Developed a fragmentation pipeline in 'src/domain_fronter.rs' to handle outbound payloads exceeding 5 MiB. Given the ~50 MiB inbound body limit on Google Apps Script, large uploads (POST/PUT) are split into sequential fragments. Each fragment is wrapped in an envelope containing a unique 'X-MHRV-Upload-ID' and sequencing headers ('X-MHRV-Chunk-Index', 'X-MHRV-Chunk-Total'). On the backend, fragments are temporarily stored in Google Drive and reassembled once the final chunk arrives, allowing the system to support massive uploads that would otherwise exceed script execution boundaries.

**- Rolling 24-Hour Quota Ledger**
Implemented a thread-safe sliding-window ledger using 'Vec<Instant>' for each script ID in 'src/domain_fronter.rs'. Unlike a fixed daily reset, this logic prunes expired timestamps older than 24 hours during every script selection event. This precisely mirrors Google's rolling quota reset cadence, ensuring the round-robin selector only routes traffic to scripts with verifiable capacity. This prevents the "thundering herd" problem where scripts are hit immediately after a hard reset while still being rate-limited by the backend.

**- Granular Failure Classification and Quarantine**
Enhanced the 'do_relay_once_with' logic to perform deep inspection of failure responses. The system now differentiates between transient network timeouts and authoritative account limits. Hard failures (HTTP 429, 403, or responses containing "Quota Exceeded") trigger a strict 24-hour quarantine window for the affected script. Transient socket errors or 5xx responses from the Google frontend trigger a brief 10-minute cooldown. This intelligent classification maximizes pool utilization by ensuring that scripts are only blacklisted for durations that match their specific failure recovery window.

**- Remote DNS Enforcement and Isolation**
Enforced SOCKS5 remote DNS resolution (Address Type 0x03) within 'src/proxy_server.rs' to eliminate domain leakage. The proxy intercepts connection attempts and passes the raw hostname directly to the encrypted tunnel, bypassing 'std::net::ToSocketAddrs' and other local resolution bindings. This ensures that destination metadata is never exposed via plaintext DNS queries to local ISP servers, maintaining full end-to-end privacy for the target hostnames.

**- System Proxy Self-Healing and Watchdog**
Established a dual-layer preservation strategy for Windows system proxy settings in 'src/main.rs'. A global panic hook using 'std::panic::set_hook' is registered to forcefully clear the 'ProxyEnable' and 'ProxyServer' registry keys during any unhandled exception. Complementing this, a boot-initialization routine flushes orphaned proxy settings from previous ungraceful exits. This self-healing architecture prevents the system's network configuration from being left in a broken state if the process is terminated via power loss or task termination.

**- WinINet System Proxy Synchronization**
Integrated direct Win32 FFI bindings to 'InternetSetOptionW' in 'src/bin/ui.rs' to ensure registry changes are propagated instantly. By broadcasting the 'INTERNET_OPTION_SETTINGS_CHANGED' and 'INTERNET_OPTION_REFRESH' flags, the OS network subsystem notifies active applications (such as Chrome, Edge, and background services) to flush their proxy caches. This provides seamless, real-time toggling of the system proxy state without requiring browser restarts or waiting for OS-level cache timeouts.

**- Local Traffic Filtering (block_hosts)**
Added a local interception gate in 'src/proxy_server.rs' that matches destination hosts against a 'block_hosts' configuration. Requests to trackers, ads, and telemetry endpoints (identified by exact match or suffix) are short-circuited with a 204 No Content response locally. This proactive filtering preserves the user's limited Apps Script execution quota for meaningful content and reduces overall latency by eliminating unnecessary remote round-trips for non-essential traffic.

**- UI Modernization and Live Progress Tracking**
Overhauled 'src/bin/ui.rs' with a high-contrast 'Obsidian' theme and real-time operational metrics. The interface now features a live 'ProgressBar' bound to the sliding-window ledger to visualize quota consumption. Status indicators were updated with sine-wave-driven alpha pulsing to provide interactive feedback on background connection states, while informational blocks were added to provide technical context on local loopback decryption and certificate sandboxing.

**- Technical Fixes and Maintenance**
Resolved a compatibility issue in 'src/main.rs' and 'src/bin/ui.rs' by migrating 'RegKey::predefined' calls to the modern 'RegKey::predef' API as required by the latest 'winreg' library. Fixed a '#[warn(unused_variables)]' warning by removing the unused variable 'n' in the 'next_script_id' implementation in 'src/domain_fronter.rs'.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Auto detect text files and perform LF normalization
+* text=auto
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2654,6 +2654,7 @@ dependencies = [
  "tun2proxy",
  "url",
  "webpki-roots 0.26.11",
+ "winreg",
  "x509-parser",
  "zstd",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,7 @@ eframe = { version = "0.28", default-features = false, features = [
     "accesskit",
 ], optional = true }
 url = "2.5.8"
+winreg = "0.55"
 
 # Unix-only deps. Must come after `[dependencies]` because starting a new
 # table here otherwise ends the main one — anything below it (incl. eframe)
--- a/src/bin/ui.rs
+++ b/src/bin/ui.rs
@@ -24,11 +24,20 @@ const WIN_HEIGHT: f32 = 680.0;
 const LOG_MAX: usize = 200;
 
 fn main() -> eframe::Result<()> {
+    // Install default rustls crypto provider (ring).
     let _ = rustls::crypto::ring::default_provider().install_default();
+
     // Re-point HOME at the invoking user if this binary was launched
     // under sudo (see cert_installer::reconcile_sudo_environment). Must
     // run before any data_dir / firefox_profile_dirs call.
     reconcile_sudo_environment();
+
+    #[cfg(target_os = "windows")]
+    {
+        // Boot-up Initialization Proxy State Flush
+        sync_wininet_proxy(false, 0);
+    }
+
     mhrv_rs::rlimit::raise_nofile_limit_best_effort();
 
     let shared = Arc::new(Shared::default());
@@ -95,7 +104,17 @@ fn main() -> eframe::Result<()> {
         "mhrv-rs",
         options,
         Box::new(move |cc| {
-            cc.egui_ctx.set_visuals(egui::Visuals::dark());
+            let mut premium_visuals = egui::Visuals::dark();
+            premium_visuals.panel_fill = egui::Color32::from_rgb(18, 20, 24);     // Deep Obsidian Canvas
+            premium_visuals.window_fill = egui::Color32::from_rgb(26, 29, 36);    // Slate Card Surface
+            premium_visuals.widgets.active.bg_fill = egui::Color32::from_rgb(59, 130, 246);  // Accent Cobalt
+            premium_visuals.widgets.hovered.bg_fill = egui::Color32::from_rgb(37, 99, 235);
+            premium_visuals.widgets.inactive.bg_fill = egui::Color32::from_rgb(31, 41, 55);
+            premium_visuals.widgets.active.rounding = egui::Rounding::same(10.0);
+            premium_visuals.widgets.hovered.rounding = egui::Rounding::same(10.0);
+            premium_visuals.widgets.inactive.rounding = egui::Rounding::same(10.0);
+            cc.egui_ctx.set_visuals(premium_visuals);
+
             Ok(Box::new(App {
                 shared,
                 cmd_tx,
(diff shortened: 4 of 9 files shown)


w0l4i (approved):
Great work, keep it going champ 👍

therealaleph (changes requested):
Thanks for putting this together, but I cannot accept this as one combined PR.

This bundles too many high-risk changes at once: quota accounting, transport security behavior, dynamic SNI selection, remote DNS/proxy behavior, Windows proxy integration, failure quarantine, `block_hosts`, UI redesign, dependency changes, and `.gitattributes` churn. Several of those are security- or transport-sensitive, and they need to be reviewed and tested independently.

Please split this into small PRs with one behavioral change each. A good first slice would be either quota accounting or one clearly bounded transport improvement, with focused tests and a short explanation of the failure mode it fixes. Also note that `main` is now on TOML config as of v1.9.34, so any config-facing change needs TOML examples and migration coverage.

Requesting changes for now so this does not accidentally move forward as an unreviewable bundle.

---

maybeknott:
> Thanks for putting this together, but I cannot accept this as one combined PR.
> 
> This bundles too many high-risk changes at once: quota accounting, transport security behavior, dynamic SNI selection, remote DNS/proxy behavior, Windows proxy integration, failure quarantine, `block_hosts`, UI redesign, dependency changes, and `.gitattributes` churn. Several of those are security- or transport-sensitive, and they need to be reviewed and tested independently.
> 
> Please split this into small PRs with one behavioral change each. A good first slice would be either quota accounting or one clearly bounded transport improvement, with focused tests and a short explanation of the failure mode it fixes. Also note that `main` is now on TOML config as of v1.9.34, so any config-facing change needs TOML examples and migration coverage.
> 
> Requesting changes for now so this does not accidentally move forward as an unreviewable bundle.
> 
> Answered via LLM, Supervised @therealaleph

Hi!
I am closing this PR. Instead, as per your request I have started creating small maintainable PRs. I would appreciate if you provide feedback for those changes sooner rather than later, so I can build more significant changes upon them.
Best,
May.

PR #1346, opened 2026-05-22
fix(domain_fronter): large download resilience, stream timeout decoupling, and compact log timestamps

CaptainMirage:
## Summary

- **Large download resilience** — range requests (`Range: bytes=N-`) are
  now routed through a parallel resume streaming path. Idempotent
  exit-node timeouts fall back to direct Apps Script instead of aborting.
  TLS `close_notify` is sent on shutdown so wget/curl can resume from
  the correct offset.

- **Stream timeout decoupling** — splits `request_timeout_secs`
  (header/connect phase, 30 s) from a new `stream_timeout_secs`
  (per-chunk body idle, default 300 s) so large transfers through Apps
  Script are no longer killed mid-body by the batch timeout firing
  during the drain phase.

- **Compact log timestamps** — replaces the noisy ISO 8601 default
  (`2026-05-22T12:56:28.992990Z`) with `[2026-05-22]-[12:56:28.9]`.
  Terminal output gets ANSI colouring (orange date, light blue time,
  gray structure); UI log panel and Android Logcat stay plain text.

Here are the log timestamp colours; they may vary slightly depending on the terminal, since they use ANSI colouring :
image

---
i also looked into fixing WebSocket support in Apps Script proxy mode, but in practice i dropped it

the main issue is that Apps Script web app requests don’t stay alive long enough for a stable long-lived WebSocket path, in my testing, they die around the ~6 minute mark, which matches the general Apps Script execution/runtime limits. I even built and tested a local ws tunnel layer for this, but the connection still could not survive past that limit reliably

that would already make the Apps Script WebSocket path fragile, but the bigger problem is what happens after the upstream socket dies. Some services (Discord especially) can behave badly here depending on how the client/library handles the close, so just letting the tunnel die and hoping reconnect works is not good enough. To make it usable, it would need a more dynamic reconnect/rebind system on top

at that point it becomes a lot of extra complexity for something that still has a built-in timeout ceiling every few minutes, while the Google Cloud path already solves the problem properly, so i dropped it

for lighter WebSocket use cases, like short-lived notifications or some live-feed scenarios, this limitation may be less painful. But for something like Discord, where long-lived stable gateway connections matter, it is too annoying to justify

### DIFF
--- a/src/android_jni.rs
+++ b/src/android_jni.rs
@@ -132,6 +132,7 @@ fn install_logging_once() {
             .with_target(false)
             .with_ansi(false)
             .with_writer(LogcatWriter)
+            .with_timer(crate::logging::CompactUtcTime)
             .try_init();
 
         let _ = rustls::crypto::ring::default_provider().install_default();
--- a/src/bin/ui.rs
+++ b/src/bin/ui.rs
@@ -294,6 +294,7 @@ struct FormState {
     auto_blacklist_window_secs: u64,
     auto_blacklist_cooldown_secs: u64,
     request_timeout_secs: u64,
+    stream_timeout_secs: u64,
     /// Optional second-hop exit node for CF-anti-bot bypass (chatgpt.com /
     /// claude.ai / grok.com / x.com). Config-only — no UI editor yet.
     /// See `assets/exit_node/` for the generic exit-node handler.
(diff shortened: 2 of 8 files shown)


CaptainMirage:
i assume you are quite busy aleph but if you dont mind could you respond a tad bit faster? i keep digging myself into a hole by editing files i have already made a PR for and i have to wait for one to be reviewed before i can push the other one, pardon me, i just want to move a tad bit quicker

many thanks my good sir

CaptainMirage:
aleph could you please turn on discussions tab on this repo so people will stop using the issues tab for discussions basically? thats what that tab is for afterall, i wanna use the issues for making issues to track properly and get reports on but its filled with people constantly asking the same things, we need a frequently asked Q&A in it too, would be lovely if you did that and or got someone else to be a maintainer too, thank you

PR #1382, opened 2026-05-24
fix(relay): harden large download streaming

maybeknott:
Builds on CaptainMirage's relay download-resilience work with a clean branch based on current main.

This PR preserves the two reliability commits from #1346 and adds the missing config-facing completion layer:

- separates relay header/connect timeout from per-body-chunk stream idle timeout;
- preserves `stream_timeout_secs` through desktop UI save round-trips;
- documents `request_timeout_secs` and `stream_timeout_secs` in TOML examples;
- updates English and Persian guides to reflect active range-aware download streaming and its remaining Apps Script quota cost;
- adds focused regression coverage for timeout parsing/defaults, JSON migration preservation, resume Range parsing, and non-zero-offset probe validation.

The branch intentionally excludes #1346's unrelated compact logging/UI timestamp changes and release artifact churn so the relay behavior can be reviewed independently.

Verification:
- `git diff --check` passes.
- Focused Cargo tests are currently blocked locally by the uncached `winnow v0.7.15` dependency when using `--offline`.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -2,6 +2,11 @@
 mode = "apps_script"
 script_id = "YOUR_APPS_SCRIPT_DEPLOYMENT_ID"
 auth_key = "CHANGE_ME_TO_A_STRONG_SECRET"
+# Response header/connect timeout for each relay request.
+request_timeout_secs = 30
+# Per-body-chunk idle timeout after headers arrive. Keep this larger than
+# request_timeout_secs so slow large downloads can continue streaming.
+stream_timeout_secs = 300
 
 [network]
 google_ip = "216.239.38.120"
--- a/config.full.example.toml
+++ b/config.full.example.toml
@@ -2,6 +2,11 @@
 mode = "full"
 script_id = "YOUR_APPS_SCRIPT_DEPLOYMENT_ID"
 auth_key = "CHANGE_ME_TO_A_STRONG_SECRET"
+# Response header/connect timeout for each relay request.
+request_timeout_secs = 30
+# Per-body-chunk idle timeout after headers arrive. Keep this larger than
+# request_timeout_secs so slow large downloads can continue streaming.
+stream_timeout_secs = 300
 
 [network]
 google_ip = "216.239.38.120"
--- a/docs/guide.fa.md
+++ b/docs/guide.fa.md
@@ -230,7 +230,8 @@ HTTP / HTTPS مثل قبل از Apps Script می‌رود (تغییری نمی�
 **محافظ‌های منابع:**
 - **حداکثر ۵۰ op** در هر بَچ — اگر سشن‌های فعال بیشتر باشند، مالتی‌پلکسر چند بَچ می‌فرستد
 - **سقف payload ۴ مگابایت** در هر بَچ — خیلی کمتر از ۵۰ مگابایت Apps Script
-- **timeout ۳۰ ثانیه** هر بَچ — مقصد کند / مرده نمی‌تواند سایر سشن‌ها را گیر بیاندازد
+- **timeout ۳۰ ثانیه برای اتصال و هدرها** در هر بَچ — مقصد کند / مرده نمی‌تواند سایر سشن‌ها را گیر بیاندازد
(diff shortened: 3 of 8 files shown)


CaptainMirage:
im sorry did you just basically pull my commits and added a config update when i already added an automatic system for the configs to update based on what values are being used in #1317 ? even worst case scenario if it cant update the config it just uses the default and there is no need to change that in 90% of cases.

there is no need to update the docs just to tell a normal user that the download now works normally, that shouldve been there in the first place and it was a bug not a feature to advertise, the docs are already packed with things that are not needed we dont need more clutter, we are not trying to confuse or scare away a new user its meant to be a guide not a world map

CaptainMirage:
just to add some context here: my concern isnt about blocking improvements, its about avoiding duplicate work. The behaviour this PR is touching is already handled automatically in my earlier PR (#1317), including sane defaults when config updates aren’t possible, so introducing a second, overlapping approach and extra docs for it could make things harder to maintain and more confusing for users in the long run. I’d prefer if we could either build on top of the existing implementation or consolidate the changes rather than having parallel solutions. and specifically in this case,the default `stream_timeout_secs` being 300 is fine and in most cases doesnt need to be changed since at that point if you are having problems with the stream it might not be the programs issue

and the UI update isnt needed, a lot of variables are missing from there, i am working on an overhauled version of the UI since it really needs one and a lot of things need to be changed around

maybeknott:
> just to add some context here: my concern isnt about blocking improvements, its about avoiding duplicate work. The behaviour this PR is touching is already handled automatically in my earlier PR (#1317), including sane defaults when config updates aren’t possible, so introducing a second, overlapping approach and extra docs for it could make things harder to maintain and more confusing for users in the long run. I’d prefer if we could either build on top of the existing implementation or consolidate the changes rather than having parallel solutions. and specifically in this case,the default `stream_timeout_secs` being 300 is fine and in most cases doesnt need to be changed since at that point if you are having problems with the stream it might not be the programs issue
> 
> and the UI update isnt needed, a lot of variables are missing from there, i am working on an overhauled version of the UI since it really needs one and a lot of things need to be changed around

That's fair.

I’ll close this PR and stop treating it as a separate source of truth. The download-resilience behavior should continue through #1346/#1317, and any useful pieces from my branch can be proposed later only as narrow follow-ups that do not overlap with your config migration or UI work.

I also agree on the UI/docs point. I’ll avoid adding user-guide text for this bug fix and won’t touch the desktop UI for `stream_timeout_secs`, since you already have broader UI work in progress and the default is sane for normal users.

Thanks for clarifying the #1317 behavior. I should have checked that path more carefully before opening this. Sorry for the frustration.

CaptainMirage:
> > just to add some context here: my concern isnt about blocking improvements, its about avoiding duplicate work. The behaviour this PR is touching is already handled automatically in my earlier PR (#1317), including sane defaults when config updates aren’t possible, so introducing a second, overlapping approach and extra docs for it could make things harder to maintain and more confusing for users in the long run. I’d prefer if we could either build on top of the existing implementation or consolidate the changes rather than having parallel solutions. and specifically in this case,the default `stream_timeout_secs` being 300 is fine and in most cases doesnt need to be changed since at that point if you are having problems with the stream it might not be the programs issue
> > and the UI update isnt needed, a lot of variables are missing from there, i am working on an overhauled version of the UI since it really needs one and a lot of things need to be changed around
> 
> That's fair.
> 
> I’ll close this PR and stop treating it as a separate source of truth. The download-resilience behavior should continue through #1346/#1317, and any useful pieces from my branch can be proposed later only as narrow follow-ups that do not overlap with your config migration or UI work.
> 
> I also agree on the UI/docs point. I’ll avoid adding user-guide text for this bug fix and won’t touch the desktop UI for `stream_timeout_secs`, since you already have broader UI work in progress and the default is sane for normal users.
> 
> Thanks for clarifying the #1317 behavior. I should have checked that path more carefully before opening this. Sorry for the frustration.

no frustration here mate dont get me wrong, the TOML move was quite a big PR i didnt mention every single detail i changed and made in the body, just wanted to say it here just so you know it already exists

PR #1388, opened 2026-05-24
feat(relay): prioritize mux dispatch and expose script health

maybeknott:
This groups the relay batching, local quota steering, failure quarantine, and script-health visibility pieces into one coherent relay behavior slice.

TunnelMux now classifies queued operations by dispatch urgency before applying the coalescing wait. Connection opens, connect-and-send opens, and non-empty TCP/UDP data operations bypass the short batching delay once already queued work has been drained. Empty polls and close notices remain batch-friendly so idle long-poll cadence and cleanup traffic can still piggyback without forcing extra Apps Script batches.

DomainFronter also keeps a local rolling 24-hour ledger per configured Apps Script deployment. Selection prunes expired observations and prefers non-blacklisted deployments whose locally observed call count is still below the free-tier steering threshold. If all healthy deployments are locally saturated, routing continues instead of creating a client-side outage, preserving compatibility with paid quotas, shared deployments, and cases where the local estimate is conservative.

Failure handling now separates hard quota/account failures from transient relay failures. HTTP 429/403 and recognizable quota-exceeded bodies quarantine the affected deployment for 24 hours. Transient relay failures such as 5xx responses use a short cooldown, while timeout strikes continue to protect against repeatedly selecting a stalled deployment.

The desktop UI exposes a read-only Script health panel with masked deployment IDs, local rolling-window usage, saturation status, cooldown expiry, failure reason, and timeout strike counts. The guide documents that this is local telemetry rather than an authoritative Google quota reading.

Validation:
- `git diff --check`
- `cargo test mux_priority --lib`
- `cargo test should_fire --lib`
- `cargo test quota --lib`
- `cargo test quarantine --lib`
- `cargo test script_health --lib`
- `cargo test compact_seconds_formatter --features ui --bin mhrv-rs-ui`


CaptainMirage:
one thing i would recommend, try to first do the task you wanna do, push the branch and then make the PR, generally makes it ezier to merge and it can cause problems in more active repos, currently the owner which is the only maintainer is sleeping for a few days so its fine in this case, just telling for later on 

specially the push thing, if you make a mistake or want to merge some commits with one another, its alot ezier to fix those problems if you havent pushed the branch yet, if you do push it the history stays and it can make the branch commits a bit messy, a fully local branch is alot eizer to change around and fix than one that is already pushed, just so you know :)

PR #1398, opened 2026-05-25
iOS: full-tunnel VPN app — UI parity, build fix, SNI pool

yyoyoian-pixel:
## Summary
iOS NetworkExtension full-tunnel VPN client (leaf + FakeIP → SOCKS5 → mhrv-rs).

- **Build fix:** produce the static lib via `cargo rustc --crate-type staticlib`. `cargo build --lib` also links the cdylib, which fails on iOS (undefined `___chkstk_darwin` under `-nodefaultlibs`), aborting the build and leaving a stale `.a` so device builds never picked up changes.
- **utun fd:** `getsockopt(UTUN_OPT_IFNAME)` fallback when the KVC keypath returns nil on newer iOS; `dup()` so leaf owns the fd.
- **leaf logging:** level `none` so leaf's tracing init doesn't panic against the already-installed global subscriber; panic hook logs payload + location.
- **UI:** full-tunnel mode, multiple deployment IDs, auth key show/hide, SNI pool editor with per-SNI TLS probe, auto-detect Google IP, QUIC/STUN/DoH toggles.
- **Config import:** `mhrv-rs://` (zlib) + JSON/TOML, Android-compatible.
- **Signing:** team read from a gitignored `Local.xcconfig` (no team ID in VCS); generated `.xcodeproj` and `build/` gitignored.
- App icon; version from `MARKETING_VERSION`; iOS README incl. TestFlight steps.

Other modes (apps_script/direct) and the SNI native probe are deferred to follow-up PRs.

## Test plan
- [ ] `cd ios && cp Local.xcconfig.example Local.xcconfig` (set team), `xcodegen generate`, build to device
- [ ] Connect VPN; verify traffic (browsing, Telegram)
- [ ] SNI pool toggles + "Test all" probe
- [ ] Config import/share round-trips with Android


therealaleph:

- `cargo test --lib` — 248 passed, 0 failed.
- `bash -n scripts/build-ios.sh` — passed.
- `plutil -lint ios/App/Info.plist ios/NetworkExtension/Info.plist` — passed.
- `ruby -e 'require "yaml"; YAML.load_file("ios/project.yml")'` — passed.
- `git diff --check` — passed.

I also attempted `cargo check --target aarch64-apple-ios --lib --profile release-ios`; this local Mac is currently using CommandLineTools rather than full Xcode, so the check stops at `xcrun: SDK "iphoneos" cannot be located`. That is a local SDK availability blocker, not a code failure from the PR.

Merging this now.

---

PR #1401, opened 2026-05-26
fix(tunnel-node): bound per-session memory to prevent OOM

yyoyoian-pixel:
## Summary

The tunnel-node `reader_task` appended upstream data into each session's `read_buf` with **no size limit**. With Apps Script RTT of 2-7s, fast upstreams (video, downloads) could push tens of MB between drains. Multiple concurrent sessions compounded until the VM was exhausted and the process was OOM-killed.

## Changes

- **`READ_BUF_CAP` (32 MB) backpressure** — `reader_task` pauses reads when the per-session buffer reaches the cap and resumes once the client drains it. Bounds per-session memory regardless of upstream throughput.
- **Conditional `last_active` bump** — TCP `data` ops now only refresh `last_active` on real uplink writes or when a drain returns downstream data. Empty long-poll batches no longer keep idle sessions alive past the 300s reaper (matches the existing `udp_data` `had_uplink` pattern).
- **`abort_all()` in reaper + batch-drain cleanup** — previously only `reader_handle.abort()` was called, leaking `udpgw_handle` on virtual (udpgw) sessions.
- **Diagnostic logging in `cleanup_task`** — logs TCP session count and total `read_buf` size every 30s so memory pressure is observable in logs.

## Validation

- `cargo test` in `tunnel-node`: **38 passed, 0 failed**.
- Deployed to production VPS: memory held at **~26%** (1.0 GB) over **21 hours / 72 GB** of traffic, where the prior build OOM-killed itself at the same age.

## Test plan

- [x] Unit tests pass (`cargo test`)
- [x] Deployed and monitored on production VPS for 21h — memory stable, no OOM
- [ ] Reviewer sanity-check on the backpressure sleep-loop interval (50ms)

### DIFF
--- a/tunnel-node/src/main.rs
+++ b/tunnel-node/src/main.rs
@@ -83,6 +83,13 @@ const UDP_QUEUE_LIMIT: usize = 256;
 /// a maximum-size IPv4 datagram without truncation.
 const UDP_RECV_BUF_BYTES: usize = 65536;
 
+/// Hard cap on each TCP session's `read_buf`. The reader_task pauses
+/// when the buffer reaches this size and resumes once the client drains
+/// it below the cap. Bounds per-session memory to ~32 MB regardless of
+/// upstream throughput — prevents OOM when fast upstreams (video,
+/// downloads) outpace Apps Script's 2-7 s polling interval.
+const READ_BUF_CAP: usize = 32 * 1024 * 1024;
+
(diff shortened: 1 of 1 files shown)


therealaleph:

- `cd tunnel-node && cargo test` — 38 passed, 0 failed.

Merging this now; the bounded per-session buffer/backpressure change is a good fix for the full-mode OOM path.

---

PR #1417, opened 2026-05-29
perf(relay+tunnel): H2/H1 connection tuning, 64KB copy buffer, dynamic coalescing, CDN header filtering

CaptainMirage:
## Summary

Four independent performance improvements targeting the `apps_script` relay
path and `full` tunnel mode. No breaking changes to config schema or wire
protocol — all new config fields have safe defaults and are backward-compatible.

---

### 1. H2/H1 connection timing tuning

**Files:** `src/domain_fronter.rs`, `src/tunnel_client.rs`

Adjusted the constants governing the HTTP/2 relay connection and H1 keepalive
pool to reduce per-request overhead:

- **H2 flow control windows**: initial stream and connection windows raised from
  4 MB to 16 MB. Eliminates the WINDOW_UPDATE round-trips that throttle
  throughput on high-bandwidth links — previously each request stalled waiting
  for window top-ups before the CDN would send more data.
- **H1 keepalive idle timeout**: reduced from 240s to 60s. Shorter interval means
  fewer stale sockets get reused just before the server closes them, cutting the
  reconnect-on-stale failure rate under the proxy.
- **H1 pool pre-warm**: opens 6 parallel H1 connections at startup instead of
  on-demand. First requests after startup no longer pay the cold TLS handshake
  cost.

Measured result on a typical Iran ISP path: throughput went from ~150 KB/s to
~600 KB/s on the `apps_script` relay path.

---

### 2. 64 KB copy buffer for bidirectional TCP relay

**Files:** `src/proxy_server.rs`

Replaced the custom 8 KB `select! { read → write }` loop used for all
transparent TCP passthrough with `copy_bidirectional_with_sizes(65536, 65536)`.
The old 8 KB buffer required 128 kernel syscall pairs per 1 MB transferred; the
new 64 KB buffer cuts that to 16 — an 8x reduction in per-byte overhead on the
passthrough path.

This affects all non-relay TCP flows: SNI-rewrite direct tunnels, SOCKS5 raw-TCP
forwarding, WebSocket connections, and the exit-node passthrough path.

---

### 3. Dynamic coalescing and O(n) upload buffer fix

**Files:** `src/tunnel_client.rs`, `src/config.rs`, `src/proxy_server.rs`

**Dynamic coalescing** (`mode = "full"` only):

`TunnelMux` now measures batch RTT using a ring buffer of 8 samples (median to
filter spikes) and adapts the coalesce window automatically:

- **Fast preset** — 50ms step / 300ms max: activates after 3 consecutive RTTs
  below 200ms. Targets broadband/fiber links.
- **Slow preset** — 150ms step / 600ms max: activates when RTT exceeds 500ms.
  Targets throttled, high-latency, or mobile links.
- **Hysteresis**: 3 consecutive fast readings are required to leave the Slow
  preset, preventing oscillation on jittery connections.

A new `network_preset` config field (`"auto"` / `"fast"` / `"slow"`) lets
operators pin the preset. `"auto"` (default) leaves adaptive behaviour enabled.
`0` values for `coalesce_step_ms` / `coalesce_max_ms` now correctly resolve to
the compiled defaults via the preset, instead of disabling coalescing.

**O(n) upload buffer fix** (`mode = "full"` only):

The buffered upload accumulator that reassembles split data ops into a single
batch body was typed as `Option<Bytes>`, requiring a full copy + concatenation
on every incoming chunk — O(n²) total allocation cost for a multi-chunk upload.
Changed to `Option<BytesMut>` with `extend_from_slice`, reducing to O(n)
amortized.

---

### 4. CDN response header noise filtering

**Files:** `src/domain_fronter.rs`, `src/config.rs`, `assets/apps_script/Code.gs`

Modern CDN stacks (Cloudflare, AWS, Fastly) attach metadata headers to every
response that carry no value through a MITM relay:

| Header | Why it's useless through a relay |
|--------|----------------------------------|
| `report-to`, `reporting-endpoints` | Browser error-reporting endpoint — no reporting agent on the relay path |
| `nel` | Network Error Logging — same, no collector |
| `alt-svc` | Advertises an HTTP/3 QUIC endpoint — browsers can't use it through a proxy |
| `server-timing` | CDN render timings for browser DevTools — relay-path latency makes these meaningless |
| `cf-ray`, `cf-cache-status` | Cloudflare internal request IDs — go nowhere useful |
| `x-amzn-requestid`, `x-amzn-trace-id` | AWS request tracing — same |
| `origin-trial` | Chrome experimental feature tokens |

On a Cloudflare-backed site these headers add 400–700 bytes of JSON per relay
response. Over a page load with 50 subresource requests that is ~25–35 KB of
wasted transfer — roughly 40–50ms at 600 KB/s — multiplied across every request.

**Rust side (user-configurable via config.toml):**

```toml
[relay]
strip_noise_response_headers = true  # default; set false to see raw headers
```

When enabled, the 12 listed header names are dropped from the parsed GAS relay
response in `parse_relay_json` before the HTTP response is forwarded to the
browser. Both the H2 and H1 relay paths are covered by a single code path.

**Code.gs side (deploy-time optimization):**

`STRIP_NOISE_RESPONSE_HEADERS = true` constant added near `DIAGNOSTIC_MODE` at
the top of `Code.gs`. When enabled, `_respHeaders()` filters the blocklist
before building the JSON payload, reducing the data that travels over the
GAS → Rust leg.

The two layers are independent: Code.gs reduces GAS → Rust bandwidth; the Rust
config controls what reaches the browser. Either can be disabled independently.

---

## Test plan

- [x] **Throughput**: browse a few pages through the proxy; observe speed vs the
  previous release baseline. Target >400 KB/s on a typical Iran ISP path.
- [x] **H2 startup**: proxy log shows `h2 connection established` and
  `h2 fast path active; h1 fallback pool pre-warmed with 6 connection(s)`.
- [x] **64KB copy / upload**: large file upload through the proxy completes
  without errors (tested: imgur image upload, YouTube range streaming).
- [x] **CDN header strip (Rust)**: with the proxy running,
  `curl -sI -x http://127.0.0.1:8085 https://discord.com | grep -iE 'report-to|nel|alt-svc|cf-ray'`
  returns nothing with `strip_noise_response_headers = true`; the headers
  reappear when set to `false`.
- [ ] **CDN header strip (Code.gs)**: redeploy `Code.gs` as a new version in
  the Apps Script dashboard; check the Executions log — JSON response payloads
  for CDN-backed sites should be ~400–700 bytes smaller per request.
- [ ] **Dynamic coalescing** (`mode = "full"` only): startup log shows
  `batch coalesce: auto mode`; after sustained low-RTT batches the log should
  show a switch to the Fast preset.
- [x] **network_preset = "fast"/"slow"**: setting the field in config.toml and
  restarting pins the coalesce windows to the correct values; startup log
  confirms the chosen preset.

### DIFF
--- a/assets/apps_script/Code.gs
+++ b/assets/apps_script/Code.gs
@@ -52,14 +52,18 @@ const DIAGNOSTIC_MODE = false;
 // alt-svc, server-timing, etc. These add 400-700 bytes of JSON per response
 // for no benefit — the relay ignores them and the browser never reads them.
 //
-// STRIP_NOISE_RESPONSE_HEADERS controls whether _respHeaders() filters them
-// before returning. Hardcoded true here for GAS-side payload reduction.
 // The primary user toggle is `strip_noise_response_headers` in config.toml
-// on the Rust client side, which drops them even if Code.gs sends them.
+// on the Rust client side. That field controls what the browser sees and
+// works regardless of this constant.
 //
-// Set to false only if you need to see raw origin headers in GAS logs.
+// STRIP_NOISE_RESPONSE_HEADERS is a separate GAS-side optimization: when
+// true, headers are stripped before building the JSON payload, reducing the
+// data sent over the GAS -> Rust leg. Set to true only for extra bandwidth
+// savings. Leave false (the default) so the Rust config is the sole control
+// and `strip_noise_response_headers = false` in config.toml shows raw headers
+// without also requiring a Code.gs redeploy.
 // ---------------------------------------------------------------------------
-const STRIP_NOISE_RESPONSE_HEADERS = true;
+const STRIP_NOISE_RESPONSE_HEADERS = false;
 
 const STRIP_RESPONSE_HEADERS = {
   "report-to": 1, "reporting-endpoints": 1,
--- a/assets/apps_script/Code.gs
+++ b/assets/apps_script/Code.gs
@@ -46,6 +46,32 @@ const AUTH_KEY = "CHANGE_ME_TO_A_STRONG_SECRET";
 // (Inspired by #365 Section 3, mhrv-rs v1.8.0+.)
 const DIAGNOSTIC_MODE = false;
 
+// ── Response header noise filtering ────────────────────────────────────────
+// CDN stacks (Cloudflare, AWS, Fastly, Google) attach metadata headers to
+// every response that are useless through a MITM relay: report-to, nel,
+// alt-svc, server-timing, etc. These add 400-700 bytes of JSON per response
+// for no benefit — the relay ignores them and the browser never reads them.
+//
+// STRIP_NOISE_RESPONSE_HEADERS controls whether _respHeaders() filters them
+// before returning. Hardcoded true here for GAS-side payload reduction.
+// The primary user toggle is `strip_noise_response_headers` in config.toml
+// on the Rust client side, which drops them even if Code.gs sends them.
+//
+// Set to false only if you need to see raw origin headers in GAS logs.
+// ---------------------------------------------------------------------------
+const STRIP_NOISE_RESPONSE_HEADERS = true;
+
+const STRIP_RESPONSE_HEADERS = {
+  "report-to": 1, "reporting-endpoints": 1,
+  "nel": 1,
+  "alt-svc": 1,
+  "server-timing": 1,
+  "origin-trial": 1,
+  "cf-ray": 1, "cf-cache-status": 1,
+  "x-amzn-requestid": 1, "x-amzn-trace-id": 1,
+  "x-request-id": 1, "x-correlation-id": 1,
+};
+
 // ── Optional Spreadsheet Cache ──────────────────────────────
 // Set to a valid Spreadsheet ID to enable response caching.
 // Leave as-is to disable caching entirely (zero overhead).
@@ -329,12 +355,20 @@ function _buildOpts(req) {
 }
 
 function _respHeaders(resp) {
+  var raw;
   try {
-    if (typeof resp.getAllHeaders === "function") {
-      return resp.getAllHeaders();
-    }
-  } catch (err) {}
-  return resp.getHeaders();
+    raw = typeof resp.getAllHeaders === "function"
+      ? resp.getAllHeaders()
+      : resp.getHeaders();
+  } catch (err) {
+    raw = {};
+  }
+  if (!STRIP_NOISE_RESPONSE_HEADERS) return raw;
+  var out = {};
+  for (var k in raw) {
+    if (!STRIP_RESPONSE_HEADERS[k.toLowerCase()]) out[k] = raw[k];
+  }
+  return out;
 }
 
 function _json(obj) {
--- a/src/bin/ui.rs
+++ b/src/bin/ui.rs
@@ -273,6 +273,12 @@ struct FormState {
     /// users edit `force_http1` directly when needed. Default false
     /// (HTTP/2 multiplexing on the relay leg active).
     force_http1: bool,
+    /// Adaptive relay-network tuning preset. Config-only for now; preserve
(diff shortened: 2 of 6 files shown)


CaptainMirage:
Checked for conflicts with #1414 (tunnel-node perf by brightening-eyes), no overlap, that one is entirely in `tunnel-node/` and this one is all client-side `src/` and Code.gs. The two together should cover both ends of the full-mode path nicely.

CaptainMirage:
also cant help but notice that github actions is failing silently because some webhooks got updated so it seems, the labels arent being applied so i checked, the PR #1414 also didnt get its label
https://github.com/therealaleph/MasterHttpRelayVPN-RUST/actions/runs/26667216524

therealaleph:
Thanks for the perf pass. I checked this locally from the PR head (`929927b`):

```text
cargo test --lib
253 passed
```

I am going to hold merge for now because I found two integration mismatches that should be fixed before this lands:

1. `network_preset = "auto"` does not appear to actually enable auto-adaptation through the normal `ProxyServer` path. `ProxyServer::new` currently converts unset `coalesce_step_ms` / `coalesce_max_ms` into `10` / `1000` before calling `TunnelMux::start()`. Then `TunnelMux::start()` treats any non-zero coalesce value as an explicit override and locks the preset, so unset/`auto` ends up as `custom (locked)` with `rtt_tracker = None`. Please either pass the raw config values into `TunnelMux::start()` or carry a separate "user explicitly set this" flag, then add a proxy-level test that proves the default/`auto` path creates an active `RttTracker`.

2. The header-filtering docs/comment say that setting `strip_noise_response_headers = false` in Rust passes all headers through regardless of what Code.gs sends. With `STRIP_NOISE_RESPONSE_HEADERS = true` in Code.gs, those headers are already removed before Rust receives the JSON envelope, so the Rust config cannot restore them. Either make the GAS-side filter opt-in/off by default and let Rust be the main user-facing control, or update the docs/comments to say raw headers require changing the Code.gs constant too.

The rest of the direction looks promising, especially the larger copy buffers and the half-close cleanup, but I would like these two fixed before we ship it to users on high-latency Full-mode paths.

---

CaptainMirage:
on it

CaptainMirage:
addressed both points:

1. the coalesce wiring bug was in `ProxyServer::new`, it was converting unset `coalesce_step_ms / coalesce_max_ms` (config value 0) to 10 / 1000 before passing them to `TunnelMux::start()`, which then saw non-zero values, set `locked = true`, and never created the `RttTracker`. Fixed by passing the raw 0 values through; `TunnelMux::start()` already handled 0 correctly. Added `is_adaptive()` to TunnelMux and a regression test (`auto_preset_default_coalesce_creates_rtt_tracker`) covering the default path, explicit "auto" preset, and explicit ms values.

2. flipped `STRIP_NOISE_RESPONSE_HEADERS` in Code.gs from true to false, GAS-side stripping is now opt-in. `strip_noise_response_headers = false` in `config.toml` is the sole user-facing control and works end-to-end without a `Code.gs` redeploy. Updated the comments in `Code.gs`, `config.rs`, and `domain_fronter.rs` to describe both layers accurately.

CaptainMirage:
also aleph no comment on the opening discussions and using issues properly?

CaptainMirage:
i wonder how long this will take

therealaleph:
Tested locally before merge at `5bc1aba`:

• `cargo test --lib`: 254 passed
• `cargo build --release`: passed
• `cargo build --bin mhrv-rs-ui --release --features ui`: passed
• `node --check < assets/apps_script/Code.gs`: passed

The UI build initially exposed missing config fields; I added and tested config-only round-tripping for `network_preset` and `strip_noise_response_headers` on this branch before approval. This focused relay/tunnel performance set is ready for the next patch release. Thanks @CaptainMirage.

---

CaptainMirage:
oh hey welcome back to the land of the living

